PR #703, opened 2013-07-10
Single (inline) pipe is mistakenly (?) interpreted as multiline indicator

polarblau:
When rendering a single pipe on the same line as the tag, Haml swallows the pipe — presumably because it’s interpreted as multiline indicator (http://haml.info/docs/yardoc/file.REFERENCE.html#multiline):

``` haml
%span |
```

is evaluated to:

``` html
<span></span>
```

In my eyes (and I understand that this might very well be just a matter of opinion) that’s incorrect or at least unexpected behavior, as this case contains only a single pipe, not _multiple_. IMO the pipe should be treated as a regular string in this case.

The attached test illustrates the issue.

### DIFF
--- a/test/engine_test.rb
+++ b/test/engine_test.rb
@@ -314,6 +314,10 @@ HTML
   foo
 HAML
   end
+  
+  def test_faux_multiline
+    assert_equal("<span>|</span>\n", render("%span |"))
+  end
 
   def test_end_of_file_multiline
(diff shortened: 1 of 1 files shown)


hcatlin:
This sounds like some behaviour we should figure out– but as the last comment is from 2014– I think this PR is effectively dead! Sorry we dropped the ball on this one!

Issue #748, opened 2014-02-21
Newlines are added to scripts with blocks in non-ugly, not in ugly


mattwildig:
If the result of a ruby method that is passed a block is included with `=` then the behaviour is different in ugly and non-ugly modes.

An example with Rails’ `link_to` helper:

``` haml
= link_to 'http://example.com/1' do
  %span Example 1

= link_to 'http://example.com/2' do
  %span Example 2
```

produces this in development (i.e with `ugly` set to false):

``` html
<a href="http://example.com/1"><span>Example 1</span>
</a>
<a href="http://example.com/2"><span>Example 2</span>
</a>
```

but in production (`ugly` true) produces:

``` html
<a href="http://example.com/1"><span>Example 1</span>
</a><a href="http://example.com/2"><span>Example 2</span>
</a>
```

This may not seem much, but depending on your styling can be significant.

A simpler, somewhat contrived, non-Rails example that more clearly shows the difference in results:

``` haml
-def foo
  - "Hello"

=foo do
  .ignored
=foo
```

produces

``` html
Hello
Hello
```

in normal mode and

``` html
HelloHello
```

in ugly mode.

When compiling the `push_script` method [adds a trailing newline in normal mode, but not in ugly mode](https://github.com/haml/haml/blob/f7c7ebc85aadf2b1144afaf438cb7c2b8df0cca0/lib/haml/compiler.rb#L400) if a block is given. If no block is given a [newline is added even in ugly mode](%28https://github.com/haml/haml/blob/f7c7ebc85aadf2b1144afaf438cb7c2b8df0cca0/lib/haml/compiler.rb#L390%29)

In normal mode the script is handled by the `format_script` method, which [strips trailing whitespace](https://github.com/haml/haml/blob/f7c7ebc85aadf2b1144afaf438cb7c2b8df0cca0/lib/haml/buffer.rb#L158). With the newline provided by `push_script` the result is always a single newline. In ugly mode `format_script` (usually) doesn’t handle it, so the result is whatever number newlines are returned from the method, which could be zero.

I think we should _always_ add a newline after the script in ugly mode, to reduce the difference between the two modes. Making that change only causes a couple of tests to fail, where they are expecting a single trailing newline but get two.

(This issue was prompted by this SO question: http://stackoverflow.com/questions/21741015/haml-render-different-html-in-two-different-stages)

jess:
[...]

PR #751, opened 2014-02-24
Newline after scripts

mattwildig:
A (possible) fix for #748. There may be some corner cases I haven’t considered in relation to this, but I think this gives better and more consistent behaviour than before.


k0kubun:
Considering https://github.com/haml/haml/pull/894, is this change still necessary? I wrote my thoughts in https://github.com/haml/haml/issues/748#issuecomment-282940426.

hcatlin:
Looks like this has been resolved with the other issue.

PR #766, opened 2014-04-01
add support for passing multiple arguments to block for list_of

helloenvoy:
I wanted to each_with_index on my enumerator for list_of but found that list_of only passes a single parameter to the block.  This adds a test and the corresponding implementation for passing multiple arguments to block of list_of.


k0kubun:
Umm, the branch has conflicts and the submitting user is deleted now. Closing :cry: ...

PR #769, opened 2014-04-14
Remove FormFor Helper

apotonick:
The `FormFor` patch in HAML is supposed to add proper indentation to the rendered form. This is already (correctly) implemented in HAML's `form_tag` patch. Since `#form_for` internally uses `#form_tag` there is no need for redundancy that introduced a bug.

This bug happened to render the form's fields (actually, _any_ form markup) before the form instead of within the tags. To make it even more complicated, this only is visible when changing the view context in Rails.

Anyway, this is fixed and tested now :grin:

### DIFF
--- a/test/helper_test.rb
+++ b/test/helper_test.rb
@@ -154,10 +154,12 @@ HAML
   end
 
   def test_form_for
+    size_attribute = ""
+    size_attribute = ' size="30"' if Rails.version >= '3.2.0' and Rails.version < '4.0.0'
     # FIXME: current HAML doesn't do proper indentation with form_for (it's the capture { output } in #form_for).
     def @base.protect_against_forgery?; false; end
     assert_equal(<<HTML, render(<<HAML, :action_view))
-<form accept-charset="UTF-8" action="foo" class="new_post" id="new_post" method="post">#{rails_form_opener}<input id="post_name" name="post[name]" type="text" />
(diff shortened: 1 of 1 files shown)


timoschilling:
Can we backport this into `4.0.8`?

tmertens:
:+1: to @timoschilling's request

opti:
:+1: for backport request

phillipoertel:
👍  for backport request. Or would a 4.1.0 stable release be possible? 

I'd like to use cells and haml together, but right now that means running either with a cells version thats 2+ years old (and beta) or a beta version of haml (4.1.0.beta.1); none of which my team would like to depend on.

What are you guys doing? If you're running haml 4.1.0.beta.1, any issues?

apotonick:
Hi @phillipoertel :smile: the common workaround is to reference the commit as documented, not beautiful but Haml seems to be unsupported ATM. Have you checked out Hamlit?

phillipoertel:
Thanks @apotonick; I saw your recommend hamlit, which looks promising. Two issues though:

- our codebase is large and 8 years old; with few automated acceptance tests. So if there are Haml/hamlit compatibility issues we're in trouble; we'd have to manually retest. What's your experience with the compatibility?  
- hamlit seems to be a one-man show.

Have you had any trouble with the Haml 4.1.0.beta1?

apotonick:
I don't use Haml.

PR #770, opened 2014-04-17
HTML escape interpolated code in filters

mattwildig:
HTML escape any interpolated code if the escape_html option is set.

I’ve used the existing `:escape_html` option rather than creating a new option. Does anyone think there should be separate options?

### DIFF
--- a/lib/haml/filters.rb
+++ b/lib/haml/filters.rb
@@ -163,7 +163,7 @@ module Haml
           if contains_interpolation?(text)
             return if options[:suppress_eval]
 
-            text = unescape_interpolation(text).gsub(/(\\+)n/) do |s|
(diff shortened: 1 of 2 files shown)


ianpurvis:
Hello, could you recommend the upgrade path for filters such as

```
:javascript
  new Page(#{javascript_params.to_json});
```

Seems I can't move forward without re-architecting my app to initialize via data attributes?

myabc:
@ianpurvis have you tried the following?

```haml
:javascript
  new Page(#{raw javascript_params.to_json});
```

ianpurvis:
@myabc Thanks! That works great. ✌️

marcandre:
For anyone stumbling here, there is an alternative solution to @myabc's: it is now possible to set `options.escape_filter_interpolations` to `false` (#984) to revert the behavior 🎉

PR #800, opened 2014-09-11
Unexpected removal of newline at the end of attribute value

t-uda:
This haml:

``` unexpected-removal-newline.haml
%input{:value => "some string\r\n"}
```

produces the following html:

``` unexpected-removal-newline.html
<input value='some string'>
```

It seems that the method `build_attributes` uses `Haml::Helpers.preserve` in order to replace `\n` by `&#x000A`. Thus, a newline at the end of an attribute value is removed because `Haml::Helpers.preserve` calls `chomp`.

So, I made a patch.
- Does not use `preserve` in `build_attributes`,
- Adds 2 related test cases.

In my environment, with both `ruby 1.9.3p194` and `ruby 2.0.0p451`, I passed to all tests. I also confirmed I could apply my patch to the `stable` branch without a conflict.

Could you please consider merging my patch?

Thanks for a nice template engine!

### DIFF
--- a/lib/haml/compiler.rb
+++ b/lib/haml/compiler.rb
@@ -450,7 +450,7 @@ END
           else
             value.to_s
           end
-        value = Haml::Helpers.preserve(escaped)
+        value = escaped.gsub(/\n/, '&#x000A;').delete("\r")
         if escape_attrs
           # We want to decide whether or not to escape quotes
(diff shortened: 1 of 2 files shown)


k0kubun:
Ah, this is resolved by https://github.com/haml/haml/pull/903. Please try Haml 5.0.0.beta.2.

PR #801, opened 2014-09-15
Blocks become more useful

Iazel:
HAML (and all the other template manager I've seen) have a big lack: blocks. Using a block should be more straightforward, and as a user, given this context:

``` ruby
class Context
  def em
    "<em>#{yield}</em>"
  end
end
```

I expect this code:

``` haml
Hello
= em do
   World
!!
```

To produce something like:

``` html
Hello
<em>World</em>
!!
```

And not:

``` html
Hello
World
<em>Hello
World</em>
!!
```

Yes, I know, for case like this we can use `capture_haml`, but this approach is really bad and cause a lot of issues. For example, if I use a [ViewModel](https://github.com/apotonick/cells) and want to use some pretty nice rails gem like `simple_form_for`, the only way is to clobber the actual context with all the method needed. However, if blocks work as expected, I could use the main `ActionView` object just like an helper container.
This is just some case, but I'm sure fixing this will give a lot more flexibility, plus I can't imagine a method that needs a block but doesn't use its content.

Hence, I've made this patch. If you like it, we could also _eradicate_ a lot of helper method that become futile, and replace all the `capture_haml` with just a `yield` and `=`. I think this should also speed up the compiled code.

**There's only 2 test that I've skipped because I don't think they are really useful and it's quite hard to green them maintaining the same HAML source** (the same result can be achieved with a small modification of it).

Hope you find this as useful as I do :)

### DIFF
--- a/lib/haml/compiler.rb
+++ b/lib/haml/compiler.rb
@@ -117,7 +117,6 @@ END
 
     def compile_silent_script
       return if @options.suppress_eval
-      push_silent(@node.value[:text])
       keyword = @node.value[:keyword]
 
       if block_given?
@@ -125,13 +124,22 @@ END
         # we want to restore them for each branch
         @node.value[:dont_indent_next_line] = @dont_indent_next_line
         @node.value[:dont_tab_up_next_text] = @dont_tab_up_next_text
-        yield
+
+        push_silent(@node.value[:text]) unless keyword == :end
+        if keyword == :capture
+          inject_internal_buffer { yield } 
+        else
+          yield
+        end
         push_silent("end", :can_suppress) unless @node.value[:dont_push_end]
+        return
       elsif keyword == "end"
         if @node.parent.children.last.equal?(@node)
           # Since this "end" is ending the block,
           # we don't need to generate an additional one
           @node.parent.value[:dont_push_end] = true
+          # but we must return the buffer content if close a capture
+          inject_buffer_return if @node.parent.value[:keyword] == :capture
         end
         # Don't restore dont_* for end because it isn't a conditional branch.
       elsif Parser::MID_BLOCK_KEYWORDS.include?(keyword)
@@ -139,6 +147,8 @@ END
         @dont_indent_next_line = @node.parent.value[:dont_indent_next_line]
         @dont_tab_up_next_text = @node.parent.value[:dont_tab_up_next_text]
       end
+
+      push_silent(@node.value[:text])
     end
 
     def compile_haml_comment; end
@@ -406,12 +416,8 @@ END
       flush_merged_text
 
       push_silent "haml_temp = #{text}"
-      @bufvar << '_tmp' 
-      push_new_buffer
-      yield
-      push_silent "#@bufvar.buffer.chomp"
+      inject_internal_buffer { yield }
       push_silent('end', :can_suppress) unless @node.value[:dont_push_end]
-      @bufvar.slice!(-4, 4)  # cut '_tmp'
 
       format_script_method = "#@bufvar.format_script(haml_temp,#{args.join(',')});"
       @precompiled << "#@bufvar.buffer << #{no_format ? "haml_temp.to_s;" : format_script_method}"
@@ -561,6 +567,18 @@ END
       else
         raise SyntaxError.new("[HAML BUG] Undefined entry in Haml::Compiler@to_merge.")
       end
+    end
+
+    def inject_internal_buffer
+      @bufvar << '_tmp' 
+      push_new_buffer
+      yield
+      inject_buffer_return unless @node.value[:dont_push_end]
+      @bufvar.slice!(-4, 4)  # cut '_tmp'
+    end
+
+    def inject_buffer_return
+      push_silent "#@bufvar.buffer"
     end
   end
 end
--- a/lib/haml/compiler.rb
+++ b/lib/haml/compiler.rb
@@ -10,6 +10,7 @@ module Haml
       @to_merge    = []
       @precompiled = ''
       @node        = nil
+      @bufvar      = '_hamlout'
     end
 
     def compile(node)
@@ -21,6 +22,10 @@ module Haml
       end
     ensure
       @node = parent
+    end
+
+    def new_buffer
+      "Haml::Buffer.new(haml_buffer, #{options.for_buffer.inspect})"
     end
 
     # The source code that is evaluated to produce the Haml document.
@@ -48,8 +53,8 @@ module Haml
       preamble = <<END.tr("\n", ';')
 begin
 extend Haml::Helpers
-_hamlout = @haml_buffer = Haml::Buffer.new(haml_buffer, #{options.for_buffer.inspect})
-_erbout = _hamlout.buffer
+#@bufvar = @haml_buffer = #{new_buffer}
+_erbout = #@bufvar.buffer
 @output_buffer = output_buffer ||= ActionView::OutputBuffer.new rescue nil
 END
       postamble = <<END.tr("\n", ';')
@@ -196,7 +201,7 @@ END
 
         push_merged_text "<#{t[:name]}", 0, !t[:nuke_outer_whitespace]
         push_generated_script(
-          "_hamlout.attributes(#{inspect_obj(t[:attributes])}, #{object_ref}#{attributes_hashes})")
+          "#@bufvar.attributes(#{inspect_obj(t[:attributes])}, #{object_ref}#{attributes_hashes})")
         concat_merged_text(
           if t[:self_closing] && @options.xhtml?
             " />#{"\n" unless t[:nuke_outer_whitespace]}"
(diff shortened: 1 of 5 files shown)


dontfidget:
Before I read this PR, I was going to propose a syntax using `do <` for automatically capturing the haml buffer, as in this example:

```
= content_tag 'p', do <
  %strong Hello World
```

I'm still not sure it's a bad idea.  To produce more readable templates as suggested by this PR, I suppose you could still swap the default behavior the capture vs output based on whether the line began with '=' and '-' and then allow the follow identical expressions:

```
= content_tag 'p', do
  %strong Hello World
```

```
- result = content_tag 'p', do <
  %strong Hello World
= result
```

But doing that swap would break some existing code.

apotonick:
Hm, but we already have to patch HAML to allow blocks, how would the additional syntax help to avoid that, @dontfidget ?

Iazel:
@dontfidget yes, I know that will break a good amount of code, that's why I proposed it for the next version of HAML. We all now that big num increase means some incompatibility with the existing code, plus I still haven't find a better way to do it ^^"
However, in most cases the update should be done by a sed like:

```shell
sed -s 's/^\(\s*\)-(.*)\sdo/\1=\2do/' */*.haml
```

Some more complex case will still break, bit I think it's the minority.

dontfidget:
@apotonick, using Rails `'capture` method without any patches, I can write

```
- result = content_tag 'p' do capture do
  %strong Hello World
- end end
= result
```
or, if you prefer curly braces to avoid `end `ned`:

```
- result = content_tag('p') { capture do
  %strong Hello World
- end }
= result
```

It's ugly, but it works.  This might be Rails specific, but maybe you can just call `capture_haml` directly for other frameworks.

@lazel, I just to be clear, you're saying that we change leading `-` to `=` in order to indicate that we want  the output to be written in-place to the output, right?  How does a statement like this get handled:

```ruby
- 2.times do |n|
  %strong= n
```

Also, my understanding is that you need to avoid doing this conversion for statements like this:

```ruby
- content = capture do
  = yield
```

I assume that's what the regex under question was trying to do with your initial proposal of automatically doing in-place output even with lines leading with '-'.

My concern is two-fold: first that it would be nice not to wait until HAML 6 to do this sort of thing fairly cleanly, and second, that it doesn't really provide a clean solution to the `times` case.  My proposal could live alongside your suggestion but could be added in a minor version since I don't think it affects any existing code.

Iazel:
@dontfidget 
```haml
- 2.times do |n|
  %strong= n
```
Because of the initial `-`, this case will not output anything, but if you replace `-` with `=`, then the output will be what you expect:
```html
<strong>0</strong><strong>1</strong>
```

Doing `- content = somefunc do` will correctly assign to `content` the result of the helper call. 
The `sed` I've posted in the previous comment, is based on the assumption that the majority of case when you use a block you also want to output something; hence replace the `-` operator with `=`. 

In the version proposed the helper call is considered an atom, so if you want to output the result just use the output form! It makes more sense to me.

That's said, your suggestion could be a good hack for now, but will become useless on the next release. Honestly, I don't think it's a good idea to introduce a new syntax for a time so short, but if the haml's team found it useful, I can integrate it on this PR :)

Iazel:
**NOTE** I've deleted my previous message because it was trash. I wasn't remember correctly, sorry.

@dontfidget 
```haml
- 2.times do |n|
  %strong= n
```
I should admit that this case is controversial. Actually this works as you expect, 2 strong tag are outputted, because the buffer injection take place only on the `compile_script`. 
Instead your second example (the assign one) would cause a mess.

The proposal of a new syntax is the way to correctly manage every scenario. I think the new operator should be at the start of the string, maybe something like `>` (recall the buffer gate to me) to indicate that the block will be buffered. This can be ok for the current version, but I think having the `=` to be buffered by default and `-` to be unbuffered, plus adding the `>-` (buffered silent) would be better. 
We can also introduce the unbuffered `=`, maybe `|=` (stream directly to the global) if you really want it.

Example 1st syntax:
```ruby
def repeat_strong(n)
  html = ''
  n.times do |m|
    html << "<strong>#{ yield(m) }</strong>"
  end
  html
end
```

```haml
# buffered otuput
>= repeat_strong(2) do |n|
  = n + 1

# Buffered capture, output nothing, assign it to `cont` instead:
>- cont = repeat_strong(2) do |n|
   = n + 1
```

Unbuffered silent & script doesn't change their behaviors:
```haml
- 2.times do |n|
  = n + 1

# this is still a mess...
= repeat_strong(2) do |n|
  = n + 1
```

---------

Examples 2nd syntax:

```haml
# buffered otuput
= repeat_strong(2) do |n|
  = n + 1

# unbuffered silent, normal behavior
- 2.times do |n|
  = n + 1

# buffered silent, assign correctly
>- cont = repeat_strong(2) do |n|
   = n + 1

# why the hell you want to do something like this?! Still a mess...
|= repeat_strong(2) do |n|
  = n + 1
```
--------

I think `|=` is really worthless and propose to **not** implement it. Obviously the send syntax is better for the everyday haml, but still can manage special cases.

I know that `>` is used at the end of a tag to indicate `remove whitespace external to the tag`, hence I don't know if it's that's good of a character, even if mnemonic.

Obviously I can implement the new syntax if it has the haml's team approval :)

dontfidget:
Adding a new special first character would be a huge breaking change, so I wouldn't push for it.  It's true that '<' does have a special meaning after a tag so maybe `<<` would be better to disambiguate it from the whitespace removal characters.  Another reason to put a character at the end is that you could have a multiline ruby expression and so you really start losing the ability to read the code.

On a slight tangent, I've never been clear on why there is even a `do` at the end of the line in haml.  It seems like indentation should suffice to indicate that a block is next.

dontfidget:
Apropos of that last bit, I was wondering what haml would look like with a heredoc syntax that let you insert multiple blocks of haml sort of like this:

```
= panel <<-NAME, title: <<-TITLE do
  %strong The panel
NAME
  %strong my title
TITLE
  %p
    my panel contents
```

where panel is some helper that wraps the content in a panel and takes a block for the content.

Iazel:
I don't think (know) heredoc is supported in HAML and multiline ruby code in presentation should always be avoided, especially in HAML, hence it's not thata big of a problem

dontfidget:
I was just suggesting that HAML *could* support a herdoc syntax.  In your example, a block is sort of overkill.  It just happens that you suggested a method that takes a block, but, for example, a lot of rails helpers can take a block or a string.  Back to your original example, given heredoc support, if your method `em` took a string instead of a block you could just write it as:

```
Hello
= em << HAML
   World
HAML
```
Don't get me wrong, blocks are useful too and look nice, but I think having generalized string generation using a heredoc syntax addresses the same problem.

Iazel:
For me, that's just too different from the normal HAML syntax, better would be:
```haml
Hello
= em
   Some kind
   %a(href="link")
      of complex
   string not suitable for inline
```

I think this more HAMLish, and would be translated in something like:
```ruby
"Hello" << em( %q{Some kind <a href="link">of complex</a> string not suitable for inline} )
```

@norman Any news for this feature?

apotonick:
Hey @Iazel I just wanted to let you know that we have a generic HAML gem now that collects fixes for things that should be done in a different way (e.g. how to maintain compatibility to Rails helpers, etc.).

We would greatly appreciate your help here, and maybe we can convince our HAML friends better when we show working implementations :wink: :wink: 

https://github.com/trailblazer/cells-haml

Iazel:
@apotonick I think that's a nice approach, I will see how we can integrate this feature in the spare time, probably this weekend :P

apotonick:
Hey @Iazel we can implement that feature in Erbse first, and then present to the Haml peeps. I am now deeply convinced that capturing must be part of the low level template engine, after spending about 1 week on fixing capture with Rails 3.2, 4.0., 4.2 in Cells, ERB, Rails, Haml, and Slim. It is a low-level concept and must be implemented there.

@hcatlin @mattwildig I am incredibly interested in getting 4.1 outta the door - any help needed? I know now everything about capturing and Rails helpers patching themselves and so on :grin: what's the blockers for a 4.1? :heart:

norman:
@apotonick The primary blocker has been, frankly, me. I've been promising to get 5.0.0 (we're bumping the major version due to the dropped support for older Rubies) out the door for a long time but have been distracted by many other things and never get to it. But from now until mid-July I'll have 20 hours a week for open source and intend to log many more hours on Haml.

I'm grateful for your continued attention to Haml - if you'd like I'd be happy to give you commit access and you can work on this in a feature branch, and make other changes as needed. Depending on timelines we may be able to get it out along with the other changes for 5.0.

apotonick:
Hi @norman, I'd be happy to help! We could move a lot of Rails patches to optional modules (for older Rails), so we can get rid of them once we drop support.

Also - I started to get into performance. Anything planned here? 

Thanks!!! :heart:

norman:
@apotonick I added you as a committer. Feel free to work on whatever you think is useful. Your expertise will be beneficial to the project and I'm really happy and grateful for whatever you have time to do. Performance changes are of course always welcome, as long as they don't come at the expense of maintainability.

If you make a change that would increase a semantic version's tiny number, feel free to commit direct to master, otherwise use PRs please.

I'd like to get a 5.0.rc.1 out this week, so I'll be trying to close as many bug issues as possible, and leave the feature ones for the next version.

k0kubun:
Current implementation is dynamically changed using Temple. So this PR's diff is out-of-date. 

Feel free to reopen this after resolving conflicts.

PR #802, opened 2014-09-17
Adding `while` and `until` to the list of start block keywords in the parser

camertron:
Currently `while` and `until` aren't identified as keywords by the haml parser.


k0kubun:
We need test cases for this. What you expect for templates including those keywords is not clear. Please add them.

And since it has conflicts and passed 2-3 years, closing for now. Please reopen this if you still have this problem.

PR #803, opened 2014-09-17
[3.1 stable] Adding while, until, and unless to the list of start block keywords in the parser

camertron:
Backporting this from #802.


k0kubun:
I'm sorry but 3.x is currently not maintained. Please use Haml 4.x instead.

Issue #832, opened 2015-02-13
#{'{'} is syntax error


eagletmt:
The following template gives a syntax error. Is it intentional?

```haml
%span #{'{'}
```

mattwildig:
It’s not intentional as such, it’s a limitation of the parser when detecting interpolated sections. It’s fairly simple and expects braces to be in matched pairs (this applies to a couple of other places in the code too). The difficulty in doing this “properly” is that you quickly find yourself needing to parse arbitrary Ruby code.

Whilst not ideal, this seems to have worked so far (I don’t think anyone has raised an issue about it). Is this something that you need to get to work? It might be possible to extend the code to handle slightly more complex situations than it already does.

eagletmt:
It looks a reasonable limitation to me :-p

Then could you add some comments about this limitation to the document?
The current document seems to say any Ruby code can be embedded.
http://haml.info/docs/yardoc/file.REFERENCE.html#ruby_interpolation_

norman:
Pull requests with doc additions are always welcome.

envygeeks:
I would think as a maintainer you could have used that time you wasted saying something as aggressive and lazy as you did and updated the docs just as quickly but that's under the assumption you know about your own software.

Point out where the docs are and I'll update them and add the limitation since I ran into this too.

norman:
@envygeeks When I posted that I was on vacation, on the beach, watching my kids and posting from my phone. Sorry to have been brusque - you're quite right that I could and should have said the same thing but in a friendlier tone. You'd also do well to heed your own advise yourself, I dare say - your comment really didn't need to resort to personal attacks to get its point across.

If you want to make the update yourself, I'd appreciate the help. The place to add the documentation is in the part of the docs that @eagletmt linked to.

envygeeks:
I wasn't being personal.  I don't make personal attacks and I apologize if you took it at as so.  That said I have created one but since I didn't know how to do it any better I kept it simple.

PR #834, opened 2015-02-18
Add a gotchas section to Interpolation.

envygeeks:
Adds `%span #{'{'}` to the gotchas section.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -1020,6 +1020,10 @@ might compile to:
         });
       //]]>
     </script>
+    
+#### Interpolation Gotchas
+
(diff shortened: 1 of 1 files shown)


envygeeks:
Ref #832

norman:
This is super passive agressive. Thanks anyway, I'll close this and add the docs myself.

envygeeks:
Passive aggressive? I was pretty direct in my saying that if somebody pointed where I would add the docs so I was being absolutely aggressive and sticking to what I said I would do.

norman:
Ok, thanks.

PR #847, opened 2015-06-13
extract erb object creation out of powerset loop

gmccreight:
Hey guys!  While profiling my small Sinatra app I found the slowest part was:

```
%self      total      self      wait     child     calls  name
5.52      0.312     0.160     0.000     0.152      128   ERB::Compiler::SimpleScanner2#scan
```

It is being called so many times because ERB.new(erb) is being called from within the powerset enumeration.

This small change eliminated my performance issue, and the haml tests continue to pass on Ruby 2.2.2


amatsuda:
@gmccreight Thanks!

Issue #851, opened 2015-07-24
Change parser into a temple based parser


kbrock:
Was looking into performance boosts around haml and while comparing to slim, I became aware of [temple](https://github.com/judofyr/temple), the parser behind it. It is quite quick.

It looks to be a pretty good abstraction and there are a number of templating languages written including 2 implementations of haml on temple: [hamlit](https://github.com/k0kubun/hamlit), [faml](https://github.com/eagletmt/faml), which may serve as a reference point in terms of what speedups can be possible.

Not sure if this is possible, or desired, but I saw @norman mention temple in in #848 .

This issue is to pull out the conversation and hear what other people were thinking. Of note, the languages seem to only use data as a hash, which he mentioned in that issue.

norman:
I've been working on this on again and off again for the better part of three years, but have never gotten the time to get something working well enough that I'm willing to share it yet. I would very much like Haml to move to Temple, it's just a matter of somebody getting the time to work on it.

I'm aware of both Hamllit and Faml - they are both interesting projects and I'd even be happy to have one of them eventually become the "official" Haml gem, there are just some incompatibilities that would have to be addressed. And of course the authors would also have to be interested in doing that.

teeparham:
Closing: resolved by #860 and #893.

kbrock:
@teeparham No way!
Congrats on the change. very cool stuff.

PR #860, opened 2015-12-05
Temple migration

k0kubun:
ref: https://github.com/haml/haml/issues/851

I migrated Haml's engine backend to Temple.
In this PR, I didn't change compiler process to temple-AST-based one and make breaking changes. But now `Haml::Parser` and `Haml::Compiler` are on Temple Engine's pipeline.
There is no feature addition or performance improvement but refactoring for future maintainability.

Since I implemented Hamlit v2 with original `Haml::Parser` in https://github.com/k0kubun/hamlit/pull/43, I know we can compile `Haml::Parser::ParseNode` into Temple AST (but not completely compatible about `Haml::Helpers`, though). This is just a first step for a whole migration.

### DIFF
--- a/lib/haml/temple_engine.rb
+++ b/lib/haml/temple_engine.rb
@@ -11,7 +11,7 @@ module Haml
       @compiler = @options.compiler_class.new(@options)
 
       @compiler.compile(parser.parse)
-      @compiler.precompiled_with_return_value
+      @compiler.precompiled
     end
   end
 
@@ -70,7 +70,7 @@ ensure
 @haml_buffer = @haml_buffer.upper if @haml_buffer
 end
 END
-      "#{preamble}#{locals_code(local_names)}#{precompiled};#{postamble}"
+      "#{preamble}#{locals_code(local_names)}#{precompiled}#{postamble}"
     end
 
     private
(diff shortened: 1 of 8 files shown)


lib/haml/temple_engine.rb
```diff
@@ -0,0 +1,113 @@
+require 'temple'
+
+module Haml
+  class TempleEngine < Temple::Engine
+    define_options(
+      :attr_wrapper         => "'",
```
teeparham: We can use 1.9 hash syntax since Haml requires ruby 1.9.

teeparham:
On first glance, this looks great to me. :clap: :clap: :clap:

lib/haml/temple_engine.rb
k0kubun: Fixed in https://github.com/k0kubun/haml/commit/57c6b43b2a18493dd1e0e93086d6e5b18f375d7b

voondo:
Thanks ! FYI, without it, I had terrible performance issues with Jruby 9.1.2.0.

norman:
Very excited to see this happening. Thanks @k0kubun for your fantastic work over the last few years on everything related to Haml!

PR #861, opened 2015-12-27
Use Ruby 2.3.0 on Travis

k0kubun:
Related to https://github.com/haml/haml/commit/1983bdc551e51703eb9671b5e25c6dadaa9f4665

While we can't use it via 2.3 somehow, we already have ruby-2.3.0 on Travis http://rubies.travis-ci.org/.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: false
 language: ruby
 cache: bundler
 rvm:
-  - 2.3
+  - 2.3.0
   - 2.2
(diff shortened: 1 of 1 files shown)


teeparham:
When Travis adds precompiled binaries for 2.3, setting `2.3` in the `.travis.yml` file will work. Setting only the minor version is nice, since it will run against the latest patch version. 

Until then, we need this. It's easy enough to change back, so I'll merge this. Thank you!

PR #862, opened 2016-01-08
Work with ruby 2.3's --enable-frozen-string-literal

jeremyevans:
These changes are the minimal ones necessary to allow Roda's specs
to pass.  There may well be other changes that are required.

### DIFF
--- a/lib/haml/compiler.rb
+++ b/lib/haml/compiler.rb
@@ -8,7 +8,7 @@ module Haml
       @options     = options
       @output_tabs = 0
       @to_merge    = []
-      @precompiled = ''
+      @precompiled = String.new
       @node        = nil
     end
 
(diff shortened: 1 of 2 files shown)


mattwildig:
Thanks for bringing this up.

> There may well be other changes that are required.

I think you’re right; even with these changes, running our tests with frozen string literals enabled I get:

    552 runs, 304 assertions, 29 failures, 326 errors, 0 skips

I got this by adding the magic comment to all our source files. Simply adding `--enable=frozen-string-literal` to `RUBYOPT` or the `ruby_opts` option  to `TestTask` results in a failure due to Bundler mutating a string. I don’t think there’s a way to enable it only for certain files without editing those files like this, unfortunately.

k0kubun:
This had some conflicts, so closed. 

Added `frozen_string_literal` pragma https://github.com/haml/haml/commit/8f207073c2e0e2bc296f18424b3fec3dfe659674. If you want to flip the false flags to true, please reopen another PR with succeeding tests.

PR #865, opened 2016-02-10
Fix title of REFERENCE Attributes section

hanachin:
# I'm cherry-picking https://github.com/haml/haml/pull/744 to fix header title of haml.info
## Before

> Attributes: `
> http://haml.info/docs/yardoc/file.REFERENCE.html#attributes_
## After this pull request merged

> Attributes: {} or ()

### DIFF
--- a/.yardopts
+++ b/.yardopts
@@ -1,7 +1,7 @@
 --charset          utf-8
 --readme           README.md
 --markup           markdown
---markup-provider  maruku
+--markup-provider  kramdown
 --template-path    yard
 --default-return   ""
 --title            "Haml Documentation"
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ gemspec
 
 group :docs do
   gem "yard", "~> 0.8.0"
-  gem "maruku"
+  gem "kramdown"
 end
 
 platform :mri do


k0kubun:
@mattwildig ping? I'm not sure whether this is acceptable for stable branch or not, mainly the part changing Gemfile.

mattwildig:
I’m not sure this is needed, but I don’t really know what the procedure is for making a new release (and managing the branches) and generating the docs. If the stable branch is moved to point at the current master when a release is made, then it will pick up these changes anyway. (Does it even make sense to have a `stable` branch that is moved in this way?)

I think we need to review and document the release procedures (and ideally make it so it isn’t the sole responsibility of one person).

k0kubun:
Your concerns are reasonable. Now @amatsuda may know the answer.

The original intention of the question from me was slightly different but I understood it by myself. The Haml documentation is definitely wrong and we need its fix. While I don't know how to apply the document, merging this PR for now.

hanachin:
I looked at the Rakefile in haml-site repository under the haml organization.
https://github.com/haml/haml-site/blob/master/Rakefile#L61
It seems generate the haml.info from stable branch, that is why I send a pull request to the stable branch.

I don't know specifically how haml.info documents are released, Sorry for confusing you.

k0kubun:
It seems that `rake sync` does `git push` in site directory. And the site directory is submodule of haml.github.io. Now the infrastructure is very obvious.

I have no push access to that repository.
@amatsuda Could you do `rake sync` at haml/haml-site?

PR #867, opened 2016-02-19
Add support for CSS class names containing the `@` sign ("commercial at", U+0040)

redoPop:
CSS conventions including [BEMIT](http://csswizardry.com/2015/08/bemit-taking-the-bem-naming-convention-a-step-further) encourage use of the `@` sign in CSS class names as part of [a suffix indicating the breakpoint at which a class takes effect.](http://csswizardry.com/2015/08/bemit-taking-the-bem-naming-convention-a-step-further#responsive-suffixes)

This PR adds Haml support for `@` within class names, so we can do:

``` haml
.o-media@md
```

Instead of having to do:

``` haml
%div{class: 'o-media@md'}
```

A passing test is included.

### DIFF
--- a/lib/haml/parser.rb
+++ b/lib/haml/parser.rb
@@ -533,7 +533,7 @@ module Haml
       attributes = {}
       return attributes if list.empty?
 
-      list.scan(/([#.])([-:_a-zA-Z0-9]+)/) do |type, property|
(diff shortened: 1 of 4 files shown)


k0kubun:
Since `@` doesn't conflict with Haml syntax other than Ruby's ivars and the naming is totally valid in CSS, I think there's no problem to have this change. Merging for BEMIT naming convention users.

PR #868, opened 2016-02-25
Make inline sass work with sprockets 3.x

### LINKED ISSUE #844: Incompatibility haml (4.0.6) -> sprockets (3.0.3) causing error 'wrong number of arguments' for inline :sass filter in haml template
I got this error migrating an existing Rails 3.2 project to Rails 4.2:

```
wrong number of arguments (3 for 1)

sprockets (3.0.3) lib/sprockets/context.rb:27:in `initialize'
haml (4.0.6) lib/haml/sass_rails_filter.rb:7:in `new'
haml (4.0.6) lib/haml/sass_rails_filter.rb:7:in `render'
```

HAML's SassRailsTemplate tries to call:

```
scope = ::Rails.application.assets.context_class.new(::Rails.application.assets, "/", "/")
```

But the Sprockets initializer was changed ( https://github.com/rails/sprockets/commit/c192d9f4320074c68786c8c1727c817d2e5c8dca ) to:

```
def initialize(input)
```

It only happens using the inline :sass filter but I could not reproduce this in a separate Rails project, so it's probably triggered only in some cases. https://github.com/haml/haml/issues/645 seems to be a similar but older incompatibility issue.

samphilipd:
- Fixes #844
- Sprockets 3 changed Sprockets::Context#initialize to take a hash
  argument
- This makes haml incompatible with inline sass for sprockets 2.x but
  since the last sprockets 2.x release was in 2014 this is probably acceptable

### DIFF
--- a/lib/haml/sass_rails_filter.rb
+++ b/lib/haml/sass_rails_filter.rb
@@ -4,7 +4,11 @@ module Haml
     # Rails's asset helpers to be used inside Haml Sass filter.
     class SassRailsTemplate < ::Sass::Rails::SassTemplate
       def render(scope=Object.new, locals={}, &block)
-        scope = ::Rails.application.assets.context_class.new(::Rails.application.assets, "/", "/")
+        scope = ::Rails.application.assets.context_class.new(
+          environment: ::Rails.application.assets,
+          filename: "/",
+          metadata: {}
+        )
         super
       end
 


shuber:
Is this good to go?

samphilipd:
@shuber I realized there is a bigger problem here...

This fix _does_ work when dynamic assets compilation is enabled. However, any inline sass will fail to compile in production for sprockets >= 3.0 and that's because the asset pipeline is completely disabled in this case. There is no easy workaround.

So at the end I'm not sure if it's the right thing to merge this because it encourages people to continue to use inline sass when in fact it's completely unsupported.

shuber:
@samphilipd Ah ok gotcha, that makes sense. Sounds like the best fix is to remove inline scss blocks from views completely. Thanks for the update!

jvenezia:
This is not working in production as ::Rails.application.assets is nil.

It's working by replacing `::Rails.application.assets` with `::Sprockets::Railtie.build_environment(Rails.application)`. But I'm not really sure if it's the ideal solution here... Would love some help of someone who understand this!

```ruby
module Haml
  module Filters
    class SassRailsTemplate
      def render(scope = Object.new, locals = {}, &block)
        environment = ::Sprockets::Railtie.build_environment(Rails.application)
        scope = environment.context_class.new environment: environment, filename: '/', metadata: {}
        super
      end
    end
  end
end
```

emaxi:
Also @jvenezia I needed to use ::Rails.application instead Rails.application

NielsKSchjoedt:
+1

k0kubun:
> This makes haml incompatible with inline sass for sprockets 2.x but
since the last sprockets 2.x release was in 2014 this is probably acceptable

I'm against this point. Even for rails v5.1.0.rc1, it supports sprockets-rails >= 2.0.0 and it means that Rails 5.1 supports Sprockets 5.1!
If you want to drop support for Sprockets 2.x, please drop it from latest Rails first.

@samphilipd Could you add branching the behavior by Sprockets version if you have a chance?

k0kubun:
Added patches from your guys. Added https://github.com/haml/haml/commit/5e451f0e29c8511035badc232395cdfaa41d23c8 too as I think it's okay.

BTW, these sources may be moved to haml-rails in the future.

Issue #306, opened 2010-10-18
class-attribute is reordered when using .foo and (class="foo")


sirlantis:
I was confused by this output when I tried to use jquery-tmpl conditions in the class:

```
# .a(class="{{if x}}{{/if}}")
Haml::Engine.new('.a(class="{{if x}}{{/if}}")').to_html
#=> "<div class='a x}}{{/if}} {{if'></div>\n"

# %div(class="a {{if x}}{{/if}}")
Haml::Engine.new('%div(class="a {{if x}}{{/if}}")').to_html
#=> "<div class='a {{if x}}{{/if}}'></div>\n"
```

The current behavior prevents duplicate classes (`.foo(class="foo")`) but it would be better just to leave the class-attribute order untouched.

k0kubun:
Closing because the decision was tentative and it passed 7 years without progress.

topherfangio:
For anyone else hitting this problem and needing a quick solution, I have hacked together some quick code (based on https://github.com/haml/haml/pull/882) that you can require to fix this.

```ruby
# Override Haml to NOT sort class attributes.
#
# This is required because Semantic/Fomantic care about class ordering :facepalm:
#
# @see https://github.com/haml/haml/issues/306
module Haml
  module AttributeBuilder
    class << self
      alias_method :original_merge_value, :merge_value

      def merge_value(key, to, from)

        if key == 'class'
          merged_class = filter_and_join(from, ' ')
          if to && merged_class
            merged_class = (to.split(' ') | merged_class.split(' ')).join(' ')
          elsif to || merged_class
            merged_class ||= to
          end
          merged_class
        else
          original_merge_value(key, to, from)
        end

      end
    end
  end
end
```

I have not seen any problems yet, but please let me know if you find anything!

hcatlin:
This has been implemented.

PR #882, opened 2016-08-26
Preserve order of class names in Haml tags

TooManyBees:
Since Ruby 1.8.7 is no longer supported, and therefore the Array `|`
method preserves the order of their elements, there should be no technical
need to sort class names of Haml tags.

In addition to likely boosting performance (one less array instantiated
per tag, though I did not run any benchmarks), this fixes weird
interactions that Haml has with libaries like Hamlbars.

Take, for instance:
  `%section.class-name{ class: "{{ maybe ? 'top' : 'bottom' }}" }`

When class names are sorted, the contents of the curly-braces
expression gets reordered so that it's no longer valid Handlebars.
By not sorting, the two sources of classes (`.class-name` and the
`class` attribute) get merged without disrupting the Handlebars.

### DIFF
--- a/lib/haml/attribute_builder.rb
+++ b/lib/haml/attribute_builder.rb
@@ -125,7 +125,7 @@ module Haml
         elsif key == 'class'
           merged_class = filter_and_join(from, ' ')
           if to && merged_class
-            merged_class = (merged_class.split(' ') | to.split(' ')).sort.join(' ')
+            merged_class = (to.split(' ') | merged_class.split(' ')).join(' ')
           elsif to || merged_class
             merged_class ||= to
           end
--- a/test/engine_test.rb
+++ b/test/engine_test.rb
@@ -153,14 +153,14 @@ class EngineTest < Haml::TestCase
 
   def test_class_attr_with_array
     assert_equal("<p class='a b'>foo</p>\n", render("%p{:class => %w[a b]} foo")) # basic
-    assert_equal("<p class='a b css'>foo</p>\n", render("%p.css{:class => %w[a b]} foo")) # merge with css
-    assert_equal("<p class='b css'>foo</p>\n", render("%p.css{:class => %w[css b]} foo")) # merge uniquely
+    assert_equal("<p class='css a b'>foo</p>\n", render("%p.css{:class => %w[a b]} foo")) # merge with css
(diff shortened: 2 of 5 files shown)


TooManyBees:
The automated tests for this will amost certainly fail because I needed to adjust the haml-spec submodule to account for classes retaining their order. I'm not sure how to submit changes to both repos together (or submit PRs to both repos separately?). I'm also wondering how this change, and the accompanying change to haml-spec, would affect other implementations (I noticed there are spec runners for Perl and Lua there).

Can you advise? I think this is a worthwhile change to Haml.

k0kubun:
Ah, this issue is for https://github.com/haml/haml/issues/306. While we had an issue in Ruby 1.8.6 as commented in https://github.com/haml/haml/issues/306#issuecomment-477245, the problem is obsolete. So personally wise, I'm +1 for this change, mainly for performance.

@TooManyBees Could you open PR to haml-spec? And in this PR, could you checkout that branch's haml-spec using some `git submodule` commands and pass tests in this PR? 

Of course fixing haml-spec should come first, but we can checkout back haml-spec submodule after merging this PR if changes for haml-spec is accepted.

k0kubun:
The comment from me is that now you need to change `Haml::AttributeBuilder.merge_value` and `Haml::AttributeCompiler#compile`.

TooManyBees:
Thanks. The haml-spec PR is https://github.com/haml/haml-spec/pull/15, and this is rebased against master.

k0kubun:
@amatsuda As gem owner, how do you think for https://github.com/haml/haml-spec/pull/15 and this PR?

I think this feature itself is good to go. But maybe we need to ship it with Haml 5.1 or Haml 6.0.

biastoch2:
Waiting desperately for this PR to get merged, since Semantic UI makes use of substring attribute selectors and the class sorting of HAML breaks those selectors :-(

hcatlin:
I have merged in the changes to the spec!

hcatlin:
D'oh, like a dork I totally duplicated this work in #1032 !

This PR is missing documentation updates and an update to the haml-spec version, so I'm going to merge this one in and will rework my PR to cover the missing elements to this PR.

PR #885, opened 2016-09-23
Fix compiler to work without deprecation on ruby 2.4

jeremyevans:
Forwardable in ruby 2.4 prints a deprecation warning if forwarding
to a private method (preview2 actually raises an exception).
Haml::Engine forwards precompiled_method_return_value to
Haml::Compiler, so the easiest way to fix this is to make
Haml::Compiler#precompiled_method_return_value public.


amatsuda:
@jeremyevans Yes, I'm also seeing this on my apps. Thank you for the fix!

PR #899, opened 2017-02-07
Add ugly mode tests

k0kubun:
related to https://github.com/haml/haml/issues/894

Even though ugly mode is used for production, currently test coverage for ugly mode is poor. So I just copied `EngineTest` as `PrettyEngineTest` and let `EngineTest` test ugly mode. I named not `UglyEngineTest` but `EngineTest` because we'll drop pretty mode in the future.

I fixed `EngineTest`'s test results to be the same as the ugly mode behavior at https://github.com/haml/haml/commit/ed361008151d14c1168a7804f7f2a0d4a414e423.

Also, I fixed unconsidered bugs which I introduced by Temple migration and are detected by these new tests.

### DIFF
--- a/lib/haml/compiler.rb
+++ b/lib/haml/compiler.rb
@@ -353,7 +353,7 @@ module Haml
 
       unless block_given?
         format_script_method = "_hamlout.format_script((#{text}\n),#{args.join(',')});"
-        push_generated_script(no_format ? "#{text}\n" : format_script_method)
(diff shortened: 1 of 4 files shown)


k0kubun:
This branch cared about engine tests. rm-branch's fixes are still absolutely necessary.

amatsuda:
👏

PR #900, opened 2017-02-07
Add ugly mode template tests

k0kubun:
Continued from https://github.com/haml/haml/pull/899. Improved the coverage of ugly mode by adding a ugly version of template tests. Then I fixed a bug I introduced in https://github.com/haml/haml/pull/899... :bow:


k0kubun:
Test results are different between my environment and Travis... Researching.

Issue #891, opened 2017-01-28
Cross-site Scripting When Outputting Params Hash


mccabe615:
During a security review of an application that uses HAML I discovered that HAML doesn't properly encode the params hash when a specific key isn't used. I've created a sample application with instructions of how to recreate the issue. It seems HAML doesn't properly encode the params hash key when the full hash is included in a template as shown below.

```
-# not properly encoded
          %input{:type=>"hidden", :data=>params, :id=>"current_tab"}
-# properly encoded
          %input{:type=>"hidden", :data=>params[:id], :id=>"current_tab"}
          %p
```

[Example app](https://github.com/mccabe615/haml_xss_example)

mattwildig:
The underlying issue here seems to be that although we [escape attribute _values_](https://github.com/haml/haml/blob/ed361008151d14c1168a7804f7f2a0d4a414e423/lib/haml/compiler.rb#L444-L451), we don’t do anything for attribute _names_.

Here’s an example:

```haml
%input{"><script>alert(1)</script>" => 'foo'}
```
which produces:
```html
<input ><script>alert(1)</script>='foo'>
```
I guess the assumption was attribute names wouldn’t be dynamically created, they would be part of the Haml source and only the values would vary.

I don’t know how we should handle this. I don’t think just HTML escaping attribute names  would be correct, as they are not “content” in the same way as text or attribute values are. The [HTML spec says this about attribute names](https://html.spec.whatwg.org/multipage/syntax.html#attributes-2):

> Attribute names must consist of one or more characters other than the space characters, U+0000 NULL, U+0022 QUOTATION MARK ("), U+0027 APOSTROPHE ('), U+003E GREATER-THAN SIGN (>), U+002F SOLIDUS (/), and U+003D EQUALS SIGN (=) characters, the control characters, and any characters that are not defined by Unicode.

The obvious solution would be to just strip out those characters from attribute names when processing attributes (perhaps issue a warning if any are found). I can’t see any way of restricting that processing to just “dynamic” attributes though, so it may involve a performance hit.

k0kubun:
> The obvious solution would be to just strip out those characters from attribute names when processing attributes (perhaps issue a warning if any are found).

I think stripped attribute name is not intentional for everybody. `params` is user's input. Thus this request should be 4XX IMO. Rendering user's input as attribute name is definitely vulnerable, which is not Haml's fault.

[...]

PR #905, opened 2017-02-14
Raise exception for unsafe attribute name

k0kubun:
close https://github.com/haml/haml/issues/891

Developers should never embed user's input as attribute name. User's input can include invalid characters defined in [HTML spec](https://html.spec.whatwg.org/multipage/syntax.html#attributes-2). So I changed Haml to raise an error for that case.
For some exceptional cases that developers want to use user's input like `params`'s keys in attribute name, they should handle this exception and respond 4XX errors. It should be 5XX for normal cases.

### DIFF
--- a/lib/haml/attribute_builder.rb
+++ b/lib/haml/attribute_builder.rb
@@ -15,11 +15,7 @@ module Haml
             data_attributes = attributes.delete(key)
             data_attributes = flatten_data_attributes(data_attributes, '', join_char)
             data_attributes = build_data_keys(data_attributes, hyphenate_data_attrs, key)
-            data_attributes.keys.each do |key|
-              if key =~ INVALID_ATTRIBUTE_NAME_REGEX
-                raise InvalidAttributeNameError.new("Invalid attribute name '#{key}' was built!")
-              end
-            end
+            verify_attribute_names!(data_attributes.keys)
             attributes = data_attributes.merge(attributes)
           end
         end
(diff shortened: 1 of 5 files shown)


lib/haml/attribute_builder.rb
```diff
@@ -12,6 +15,11 @@ def build_attributes(is_html, attr_wrapper, escape_attrs, hyphenate_data_attrs,
             data_attributes = attributes.delete(key)
             data_attributes = flatten_data_attributes(data_attributes, '', join_char)
             data_attributes = build_data_keys(data_attributes, hyphenate_data_attrs, key)
+            data_attributes.keys.each do |key|
```
k0kubun: > https://github.com/haml/haml/issues/891#issuecomment-276227989
> I can’t see any way of restricting that processing to just “dynamic” attributes though, so it may involve a performance hit.

This is the way that restricts processing to just dynamic attributes.

k0kubun:
Also, I unified the behavior for the static attributes. https://github.com/haml/haml/pull/905/commits/11738e71591a419c9525cf1acac4250ca87c5f4e
Even if developers write them directly, it's invalid as HTML.

lib/haml/attribute_builder.rb
```diff
@@ -12,6 +15,7 @@ def build_attributes(is_html, attr_wrapper, escape_attrs, hyphenate_data_attrs,
             data_attributes = attributes.delete(key)
             data_attributes = flatten_data_attributes(data_attributes, '', join_char)
             data_attributes = build_data_keys(data_attributes, hyphenate_data_attrs, key)
+            verify_attribute_names!(data_attributes.keys)
```
k0kubun: > https://github.com/haml/haml/issues/891#issuecomment-276227989
> I can’t see any way of restricting that processing to just “dynamic” attributes though, so it may involve a performance hit.

This is the way that restricts processing to just dynamic attributes.

mattwildig:
This looks good to me. Raising an exception makes a lot sense as the intended HTML would be obviously wrong,  and now that we can analyse the attributes at compile time we can treat all names the same way without a big performance hit.

amatsuda:
@k0kubun Thank you!

Issue #883, opened 2016-09-07
Circular require warning when loading under Rails 5


ijdickinson:
I have a Rails engine project that has been running fine under 4.2.3. Attempting to upgrade to Rails 5.0.0.1 today, I get a warning when I run my tests (see long stacktrace below).  What I'm not sure about is whether this was always occurring, and Rails 5 is now more picky about reporting the circularity, or whether it's a change introduced by Rails 5.

My tests still complete, but the warning makes for noisy log files.

Stack trace:

```
/home/ian/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/activesupport-5.0.0.1/lib/active_support/dependencies.rb:293: warning: loading in progress, circular require considered harmful - /home/ian/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/haml-4.0.7/lib/haml/template.rb
from /home/ian/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/rake-11.2.2/lib/rake/rake_test_loader.rb:4:in  `<main>'
from /home/ian/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/rake-11.2.2/lib/rake/rake_test_loader.rb:4:in  `select'
from /home/ian/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/rake-11.2.2/lib/rake/rake_test_loader.rb:9:in  `block in <main>'
from /home/ian/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/rake-11.2.2/lib/rake/rake_test_loader.rb:9:in  `each'
from /home/ian/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/rake-11.2.2/lib/rake/rake_test_loader.rb:10:in  `block (2 levels) in <main>'
from /home/ian/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/rake-11.2.2/lib/rake/rake_test_loader.rb:10:in  `require'
from /home/ian/workspace/epimorphics/qonsole-rails/test/controllers/qonsole_rails/qonsole_controller_test.rb:1:in  `<top (required)>'
from /home/ian/workspace/epimorphics/qonsole-rails/test/controllers/qonsole_rails/qonsole_controller_test.rb:1:in  `require'
from /home/ian/workspace/epimorphics/qonsole-rails/test/test_helper.rb:23:in  `<top (required)>'
[...]

PR #908, opened 2017-03-06
Do not load modules early when in rails

lukaso:
Early loading causes a conflict with rspec and view specs.

See issue to reproduce: https://github.com/haml/haml/issues/883

I tried to ensure all the modules are loaded using the `ActiveSupport.on_load` mechanism to avoid further problems. Because I'm not sure all the situations that haml is used in, I did a fairly robust (conservative) approach.

This fix should be forward ported to `haml5`.

### DIFF
--- a/lib/haml/railtie.rb
+++ b/lib/haml/railtie.rb
@@ -1,41 +1,19 @@
-if defined?(ActiveSupport)
-  require 'haml/template/options'
-  ActiveSupport.on_load(:before_initialize) do
-    ActiveSupport.on_load(:action_view) do
-      require "haml/template"
-    end
-  end
-end
+require 'haml/template/options'
 
 module Haml
   class Railtie < ::Rails::Railtie
     initializer :haml do |app|
-      if defined?(ActiveSupport)
-        ActiveSupport.on_load(:before_initialize) do
-          ActiveSupport.on_load(:action_view) do
-            if defined?(::Sass::Rails::SassTemplate) && app.config.assets.enabled
-              require "haml/sass_rails_filter"
-            end
-          end
-        end
-      else
+
+      ActiveSupport.on_load(:action_view) do
         require "haml/template"
+
         if defined?(::Sass::Rails::SassTemplate) && app.config.assets.enabled
           require "haml/sass_rails_filter"
         end
+
+        require "haml/helpers/safe_erubis_template"
+        Haml::Filters::Erb.template_class = Haml::SafeErubisTemplate
       end
     end
   end
-end
-
-if defined?(ActiveSupport)
-  ActiveSupport.on_load(:before_initialize) do
-    ActiveSupport.on_load(:action_view) do
-      require "haml/helpers/safe_erubis_template"
-      Haml::Filters::Erb.template_class = Haml::SafeErubisTemplate
-    end
-  end
-else
-  require "haml/helpers/safe_erubis_template"
-  Haml::Filters::Erb.template_class = Haml::SafeErubisTemplate
 end
--- a/lib/haml/railtie.rb
+++ b/lib/haml/railtie.rb
@@ -10,13 +10,33 @@ end
 module Haml
   class Railtie < ::Rails::Railtie
     initializer :haml do |app|
-      require "haml/template"
-      if defined?(::Sass::Rails::SassTemplate) && app.config.assets.enabled
-        require "haml/sass_rails_filter"
+      if defined?(ActiveSupport)
+        ActiveSupport.on_load(:before_initialize) do
+          ActiveSupport.on_load(:action_view) do
+            if defined?(::Sass::Rails::SassTemplate) && app.config.assets.enabled
+              require "haml/sass_rails_filter"
+            end
+          end
+        end
+      else
+        require "haml/template"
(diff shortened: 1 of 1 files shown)


lib/haml/railtie.rb
```diff
@@ -10,13 +10,32 @@
 module Haml
   class Railtie < ::Rails::Railtie
     initializer :haml do |app|
-      require "haml/template"
-      if defined?(::Sass::Rails::SassTemplate) && app.config.assets.enabled
-        require "haml/sass_rails_filter"
+      if defined?(ActiveSupport)
```
k0kubun: Are you sure this check is required? I think this always returns truthy value because Railtie depends on ActiveSupport.

lib/haml/railtie.rb
```diff
@@ -10,13 +10,32 @@
...
+        ActiveSupport.on_load(:before_initialize) do
```
k0kubun: While I'm not sure whether we should use `before_initialize` or not, there is no need to configure before_initialize hook inside initializer, isn't it? I prefer something like:

```rb
module Haml
  class Railtie < ::Rails::Railtie
    config.before_initialize do
      # ...
    end
  end
end
```

lib/haml/railtie.rb
```diff
@@ -10,13 +10,32 @@
...
+          ActiveSupport.on_load(:action_view) do
```
k0kubun: I guess we can remove either `on_load(:before_initialize)` or `on_load(:action_view)`. Why are both of them required here?

lib/haml/railtie.rb
```diff
@@ -10,13 +10,32 @@
...
     end
   end
 end
 
-require "haml/helpers/safe_erubis_template"
-Haml::Filters::Erb.template_class = Haml::SafeErubisTemplate
\ No newline at end of file
+if defined?(ActiveSupport)
+  ActiveSupport.on_load(:before_initialize) do
+    ActiveSupport.on_load(:action_view) do
```
k0kubun: The same as https://github.com/haml/haml/pull/908/files#r104402912.

lib/haml/railtie.rb
k0kubun: > fairly robust (conservative) 

Being too conservative makes hard to understand why it is implemented as such. Since you have a reproductive repo, you can confirm which is unnecessary. Please check it and remove it.

lib/haml/railtie.rb
lukaso: I just left the checks in place, but agree. Not sure why we need to check for that given this is a railtie and the whole file shouldn't be loaded unless we are in rails. It's also in place before I touched this code and in haml 5.0.0.beta2

lib/haml/railtie.rb
lukaso: I think the 'on_load(:action_view) is the important bit. But if we can get rid of the `defined?(ActiveSupport)` then a lot can be simplified.

lib/haml/railtie.rb
```diff
@@ -1,22 +1,19 @@
...
+require 'haml/template/options'
 
 module Haml
   class Railtie < ::Rails::Railtie
     initializer :haml do |app|
-      require "haml/template"
-      if defined?(::Sass::Rails::SassTemplate) && app.config.assets.enabled
-        require "haml/sass_rails_filter"
+
+      ActiveSupport.on_load(:action_view) do
```
k0kubun: Why is on_load hook inside initializer necessary? I guess either of on_load or initializer will be called first, then the other will be called. If they are always in the same order, either of them can be removed.

lukaso: This is the recommended approach as documented here: http://pothibo.com/2015/4/customize-rails-view-controller-with-railtie and here http://blog.carbonfive.com/2011/02/25/configure-your-gem-the-rails-way-with-railtie/

In any case, the important bit is the `on_load`, otherwise the bug is not fixed. Currently the initializer runs too early.

k0kubun: > In any case, the important bit is the on_load, otherwise the bug is not fixed. Currently the initializer runs too early.

If so, I think calling `ActiveSupport.on_load` outside `initializer "haml"` is enough. Why double hooks are required? I don't think recommended by 3rd parti blog is a good reason to do that.

If the intention of `initializer "haml"` is ensuring that `ActiveSupport` is loaded before executing this `ActiveSupport.on_load` and it's really necessary, I think this PR is correct 👍 .

lukaso: Here's a concrete example from inside rails. I think the initializer is required to ensure that as you say, `ActiveSupport` is available.

https://github.com/rails/rails/blob/fe1f4b2ad56f010a4e9b93d547d63a15953d9dc2/activejob/lib/active_job/railtie.rb

lukaso:
@k0kubun Just want to make sure I've done everything needed and this is ready to merge. Thanks.

k0kubun:
> Just want to make sure I've done everything needed and this is ready to merge. Thanks.

Because this is targeted to not master but stable, we need some confidence to make the change. Since this is a little complex problem and I'm not 100% sure about what haml needs to do in Rails initializer and that the code in this PR will work about them, I can't merge this without testing by myself but I don't have time enough to do that for now.

For this, I want some comments from release owner of Haml and Rails committer. @amatsuda

PR #913, opened 2017-04-27
Docs: Help people upgrade to 5 by classifying changes

jaredbeck:
By identifying breaking changes in the changelog, it will be much easier for people to upgrade to 5.

Not being a HAML expert, I may have missed some breaking changes, but this should be a good start.


k0kubun:
Thanks! I added some changes https://github.com/haml/haml/commit/395e42cc6a8227fd90309e4d98b1031fc4ff2591 to your good start.

Some of "Fixes" are actually breaking changes, but I left some of them if it's just unexpected.

PR #914, opened 2017-04-28
Use Erubi/Erubis only in Rails template handler

k0kubun:
## Summary
To avoid changing template handler globally, I added `:filters` option to override filter compiler with Rails-specific template handler. With this change, Erubi/Erubis handler in ActionView is never used for non-Rails apps.

## Effects for existing apps
You may notice "test/filters_test.rb" is changed. But such change may occur only in the situation both Rails and non-Rails app rendering Haml templates exist (in other words, they will have breaking change. However, duplicating last newline may not affect a behavior on browser). Thus in pure Sinatra environment, there'll be no breaking change.

While Erubi and Erubis renders templates slightly differently for last newline, I think that's Rails' breaking change and not Haml's one.

### DIFF
--- a/lib/haml/helpers/safe_erubi_template.rb
+++ b/lib/haml/helpers/safe_erubi_template.rb
@@ -15,13 +15,5 @@ module Haml
       @options.merge! engine_class: Haml::ErubiTemplateHandler
       super
     end
-
-    def precompiled_preamble(locals)
-      [super, "@output_buffer = ActionView::OutputBuffer.new;"].join("\n")
-    end
-
-    def precompiled_postamble(locals)
-      [super, '@output_buffer.to_s'].join("\n")
-    end
   end
 end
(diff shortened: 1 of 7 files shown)


lib/haml/helpers/safe_erubi_template.rb
```diff
@@ -15,13 +15,5 @@ def prepare
       @options.merge! engine_class: Haml::ErubiTemplateHandler
       super
     end
-
-    def precompiled_preamble(locals)
-      [super, "@output_buffer = ActionView::OutputBuffer.new;"].join("\n")
-    end
-
-    def precompiled_postamble(locals)
-      [super, '@output_buffer.to_s'].join("\n")
```
k0kubun: See commit message written in https://github.com/haml/haml/pull/914/commits/596485a7d274ba63ebe23da73058deb76f320d16. These preamble and postamble should be removed for Erubi.

PR #918, opened 2017-05-01
Docs: Clarify changes to escaping of quotes

jaredbeck:
An example will help people (like me) understand this breaking change.

I think that JSON is a common place where this breaking change will affect people. It caught me. That's why I chose JSON for the example, but maybe a simpler example would be better?

[ci skip]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,14 @@ Breaking Changes
 * Drop dynamic quotes support and always escape `'` for `escape_html`/`escape_attrs` instead.
   Also, escaped results are slightly changed and always unified to the same characters. (Takashi Kokubun)
 * Don't preserve newlines in attributes. (Takashi Kokubun)
+* HTML escape interpolated code in filters.
+  [#770](https://github.com/haml/haml/pull/770)
+  (Matt Wildig)
+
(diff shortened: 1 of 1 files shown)


k0kubun:
I couldn't tell from `Input: #{JSON.generate(foo: "bar")}` whether it's attribute value or script interpolated in plain text. Could you write full Haml literal of attribute?

jaredbeck:
> .. whether it's attribute value or script interpolated in plain text.  Could you write full Haml literal of attribute?

I don't know what this means, sorry.  I think you're probably more qualified to write this example than me. :)

k0kubun:
Ah okay, you're mentioning a different change.

> Escape Ruby script interpolated in filters. (Matt Wildig)

It is https://github.com/haml/haml/pull/770. What you think is not related with attributes. Please move the section.

jaredbeck:
Thanks @k0kubun. Is this better?

Interesting that #770 was merged three years ago. You have been working on haml 5 for a long time! Congratulations :)

k0kubun:
> Interesting that #770 was merged three years ago. You have been working on haml 5 for a long time! 

Yes, indeed.

PR #929, opened 2017-05-16
Fix attempt to modify frozen string

jeremyevans:
Found by Roda's specs.


k0kubun:
I want to remove the root cause of this problem. But merging to use your test case.

PR #930, opened 2017-05-21
Usage of double colon when referencing Rails

emaxi:
This is related to https://github.com/haml/haml/pull/868

I got next error when using :sass filter:
 NoMethodError - undefined method `application' for Haml::Rails:Module.

Just adding double colon to Rails.application solved the error to me.


k0kubun:
Ah, `Haml::Rails` module is created by haml-rails and it would be broken.

Thanks for the patch!

aselder:
@k0kubun Can we get a 5.0.2 release so this fix can go out?

k0kubun:
I'm not a release manager. ping @amatsuda

amatsuda:
OK. I'll publish 5.0.2 within a day or so.

amatsuda:
@aselder @k0kubun Sorry for the delay. Just released 5.0.2 gem with this change. Thanks!

PR #933, opened 2017-05-24
Fix list_of helper with multi-line content

zetaben:
Hello,

The current `list_of` helper in haml 5.0.1 seem to have an issue when running in rails (at least in 5.0.0.1) when inserting multi-line content. 

In my rails app when I do a block `link_to` in the `list_of`, the content of the list is removed:
```
= list_of array do |type|               
  = link_to helper_url(type) do            
    = myhellper(type)                                         
    = type.humanize 
```

It seem to be linked to the use of `strip!` instead of `strip` the `!` version will actually return [nil if does nothing](https://ruby-doc.org/core-2.2.0/String.html#method-i-strip-21).

I managed to add a test that seem to replicate the issue without rails. It could also probably be fixed using an `result.strip! || result` which would not copy `result`.

Thanks again for HAML and saving me from ERB !

### DIFF
--- a/lib/haml/helpers.rb
+++ b/lib/haml/helpers.rb
@@ -206,7 +206,7 @@ MESSAGE
 
         if result.count("\n") > 1
           result.gsub!("\n", "\n  ")
-          result = "\n  #{result.strip!}\n"
+          result = "\n  #{result.strip}\n"
         else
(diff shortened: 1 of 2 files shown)


k0kubun:
:+1:

PR #941, opened 2017-07-11
Fix textareas only after preserved timing

k0kubun:
## Description
We call `Haml::Buffer#fix_textareas!` to rollback textarea content's first `&#x000A;`, which is unexpectedly preserved from `\n` by `Haml::Helpers#preserve`, to `\n` for every script.

But it means that we need to call `Haml::Buffer#fix_textareas!` only after we call `preserve`. So I fixed to do so.

## Benchmark
With Ruby 2.4.0 and [k0kubun/haml_bench/templates/slim_bench.haml](https://github.com/k0kubun/haml_bench/blob/833f04306bc138261d8d2102a4f0cceb3a78eca7/templates/slim_bench.haml),

### before
```
          haml 4.0.7     41.534k (± 1.4%) i/s -    210.652k in   5.072717s
          haml 5.0.1    138.067k (± 1.5%) i/s -    700.872k in   5.077348s

Comparison:
          haml 5.0.1:   138067.3 i/s
          haml 4.0.7:    41534.1 i/s - 3.32x  slower
```

### after
```
          haml 4.0.7     39.466k (± 2.1%) i/s -    199.206k in   5.049870s
          haml 5.0.1    148.617k (± 1.9%) i/s -    750.109k in   5.049201s

Comparison:
          haml 5.0.1:   148616.6 i/s
          haml 4.0.7:    39466.1 i/s - 3.77x  slower
```

Now `_hamlout`'s dependency is only `_hamlout.attributes` call for Hash in that benchmark template. It means that we have new optimization chance to lazily initialize `Haml::Buffer` instance, which is the current bottleneck of Haml.

### DIFF
--- a/lib/haml/buffer.rb
+++ b/lib/haml/buffer.rb
@@ -159,7 +159,7 @@ module Haml
     # @since Haml 4.0.1
     # @private
     def fix_textareas!(input)
-      return input unless toplevel? && input.include?('<textarea'.freeze)
+      return input unless input.include?('<textarea'.freeze)
 
       pattern = /<(textarea)([^>]*)>(\n|&#x000A;)(.*?)<\/textarea>/im
       input.gsub!(pattern) do |s|
--- a/lib/haml/compiler.rb
+++ b/lib/haml/compiler.rb
@@ -287,11 +287,11 @@ module Haml
         text = "(#{text}).strip"
       end
       if opts[:preserve_tag]
-        text = "::Haml::Helpers.preserve(#{text})"
+        text = "_hamlout.fix_textareas!(::Haml::Helpers.preserve(#{text}))"
       elsif opts[:preserve_script]
-        text = "::Haml::Helpers.find_and_preserve(#{text}, _hamlout.options[:preserve])"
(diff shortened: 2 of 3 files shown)


lib/haml/buffer.rb
```diff
@@ -159,7 +159,7 @@ def rstrip!
     # @since Haml 4.0.1
     # @private
     def fix_textareas!(input)
-      return input unless toplevel? && input.include?('<textarea'.freeze)
```
k0kubun: Note that `toplevel?` is considered to be added just for optimization, and it must be removed now to apply this in partial view (we didn't need to call in partial before because outside layout will render partial as ruby script and call `fix_textareas!` for it). 

And this fix won't cause regression as this method is idempotent and safe to call multiple times.

k0kubun:
As the original problem is heavily tested for both Rails and non-Rails templates and tests pass, I think this is safe to merge.

@haml/committers Thoughts?

k0kubun:
Since this patch does not introduce any behavior change, merging this for now. Please fell free to add some feedback later.

PR #948, opened 2017-08-08
Fix Erubi on Rails < 5.1

### LINKED ISSUE #946: Erubi superclass mismatch
### Error description
Attempting to run a Rails application with `better_errors` 2.3.0 which has a dependency to `erubi` (instead of erubis as in 2.2.0) results in the following error
```
.bundle/ruby/2.4.0/gems/haml-5.0.2/lib/haml/helpers/safe_erubi_template.rb:5: warning: toplevel constant Erubi referenced by ActionView::Template::Handlers::ERB::Erubi
rails aborted!
TypeError: superclass must be a Class (Module given)
.bundle/ruby/2.4.0/gems/haml-5.0.2/lib/haml/helpers/safe_erubi_template.rb:5:in `<module:Haml>'
.bundle/ruby/2.4.0/gems/haml-5.0.2/lib/haml/helpers/safe_erubi_template.rb:4:in `<top (required)>'
.bundle/ruby/2.4.0/gems/activesupport-5.0.5/lib/active_support/dependencies.rb:293:in `require'
.bundle/ruby/2.4.0/gems/activesupport-5.0.5/lib/active_support/dependencies.rb:293:in `block in require'
```

### Configuration
- rails 5.0.5
- haml 5.0.2
- haml-rails 1.0.0
- better_errors 2.3.0

RobinDaugherty:
ActionView prior to 5.1 does not support Erubi. haml was looking for existence of `Erubi` in the bundle, and then assuming that ActionView would support it.

This checks directly for ActionView to support Erubi rather than only looking to see if Erubi is in the bundle at all. In effect, this causes haml to use Erubis in Rails prior to 5.1.

This change is necessary because `Haml::ErubiTemplateHandler` inherits from `ActionView::Template::Handlers::ERB::Erubi`, which in Rails prior to 5.1 is actually a reference to the `Erubi` _module_, not actually a useable class. A class inheriting from a module causes a fatal exception (which can be seen in #946).

I tested this with Rails 5.0 with and without erubi in the bundle, and with Rails 5.1 with and without erubis in the bundle. (I mention this because I don't think that the CI matrix includes erubi and erubis bundle variants.)

Fixes #946.

### DIFF
--- a/lib/haml/railtie.rb
+++ b/lib/haml/railtie.rb
@@ -26,7 +26,7 @@ module Haml
           require "haml/sass_rails_filter"
         end
 
-        if const_defined?('ActionView::Template::Handlers::ERB::Erubi')
+        if defined? ActionView::Template::Handlers::ERB::Erubi
           require "haml/helpers/safe_erubi_template"
           Haml::Filters::RailsErb.template_class = Haml::SafeErubiTemplate
         else
--- a/lib/haml/railtie.rb
+++ b/lib/haml/railtie.rb
@@ -26,7 +26,7 @@ module Haml
           require "haml/sass_rails_filter"
         end
 
-        if defined?(::Erubi) && const_defined?('ActionView::Template::Handlers::ERB::Erubi')
+        if const_defined?('ActionView::Template::Handlers::ERB::Erubi')
           require "haml/helpers/safe_erubi_template"
           Haml::Filters::RailsErb.template_class = Haml::SafeErubiTemplate
         else
--- a/lib/haml/railtie.rb
+++ b/lib/haml/railtie.rb
@@ -26,7 +26,7 @@ module Haml
           require "haml/sass_rails_filter"
         end
 
-        if defined? Erubi
+        if defined?(::Erubi) && const_defined?('ActionView::Template::Handlers::ERB::Erubi')
           require "haml/helpers/safe_erubi_template"
           Haml::Filters::RailsErb.template_class = Haml::SafeErubiTemplate
         else


lib/haml/railtie.rb
```diff
@@ -26,7 +26,7 @@ class Railtie < ::Rails::Railtie
           require "haml/sass_rails_filter"
         end
 
-        if defined? Erubi
+        if defined?(::Erubi) && const_defined?('ActionView::Template::Handlers::ERB::Erubi')
```
k0kubun: I prefer having only `const_defined?('ActionView::Template::Handlers::ERB::Erubi')`. safe_erubi_template.rb does not depend on the existence of `::Erubi`.

k0kubun: Ah, `Tilt::ErubiTemplate`...

RobinDaugherty:
I agree @k0kubun. I've made that change and I can confirm that it still works.

k0kubun:
Please check travis failure. As long as I remember, `const_defined?` in Ruby 2.0.0 has bug. So could you try `defined?` too?

RobinDaugherty:
I'll test with that change as soon as I can, but it'll be a few more days.

I was trying to automate the test by adding an alternate gemfile including eribi that could be tested on Travis, so that I wouldn't have to manually test. But it seems that the gems in the gemfile are not "require"d by Bundler at startup, so adding erubi to the bundle doesn't cause `Erubi` to be defined. Any ideas on why the gem load isn't working as expected?

k0kubun:
> But it seems that the gems in the gemfile are not "require"d by Bundler at startup

Are you mentioning about testing on travis of this repository? Unlike Rails app, we don't call `Bundler.require` at startup. We would want integration test with Rails but that should be not in haml repository but haml-rails repository ideally. As haml includes many Rails integrations for now, considering backward compatibility too, we have no perfect way to solve this issue though.

k0kubun:
Thanks for fixing this. Great job 👍

RobinDaugherty:
@k0kubun I was just in the process of testing this. The change to use `defined?` instead of `const_defined?` causes the issue to come back. I'll open a PR shortly with CI changes that will show this failure, and another PR to attempt to fix this.

PR #951, opened 2017-08-17
Require entire bundle in tests, add test case for Rails 5.0 with Erubi

RobinDaugherty:
We found an issue, fixed in #948, that was caused by Erubi being loaded. The way to reproduce this was to require erubi while Rails 5.0 is in use.

I wanted to add test coverage for this, so I added a gemfile that includes rails 5.0 and erubi. This didn't reproduce the issue though, because in test_helper, only specific gems are required directly. Since `Bundler.require` is not used, only dependencies that are directly required will be available in the tests.

There seem to be two options:
1. Add `require` calls for erubi and rescue from failure.
2. Add `Bundler.require` to test_helper and allow it to load everything that's included in each gemfile.

I advocate 2 because it is closest to the way Rails works. I assume that the long-term goal is to remove Rails from this project completely in favor of haml-rails, but until it's removed, it seems to make sense to test with an environment as close to Rails as possible. 

Anyway, by making this change, I was able to [show the failure caused by having Erubi in the bundle with Rails 5.0](https://travis-ci.org/RobinDaugherty/haml/builds/265407307).

(The erubi-variant of the gemfile is currently an allowed failure, since it's not yet fixed.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ rvm:
 gemfile:
   - test/gemfiles/Gemfile.rails-5.1.x
   - test/gemfiles/Gemfile.rails-5.0.x
+  - test/gemfiles/Gemfile.rails-5.0.x.erubi
   - test/gemfiles/Gemfile.rails-4.2.x
   - test/gemfiles/Gemfile.rails-4.1.x
   - test/gemfiles/Gemfile.rails-4.0.x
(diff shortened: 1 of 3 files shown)


k0kubun:
Okay, if we could fail CI first and then fix it, I'll merge it. Sorry for previous merging. Please notify me when CI succeeds and it's okay to merge too.

RobinDaugherty:
Cool, I just added to the description...the erubi gemfile is on the allowed failures list, I'll add it back in my next PR.

RobinDaugherty:
@k0kubun CI finally passed!

k0kubun:
Ah, allowed_failures. Having large test matrix would increase the maintenance cost, so I may remove them at some moment (like Erubis is removed from all supported Rails). Anyway, I agree that this is good to have now.

Okay, let's move this ahead!

PR #952, opened 2017-08-17
Re-fix Erubi on Rails < 5.1

RobinDaugherty:
This was attempted in #948, but we made a small change after the PR was first opened. Turns out we must use check for Erubi using `defined?` and Rails' Erubi object using `const_defined?`.

If we call `defined?` with Rails's object, it gets defined by Rails. Calling `const_defined?` with the constant as a string seems to be the only way for us to see if it's been defined without _causing it to be defined_.

Without the `defined?` in front of the `const_defined?` call, Rails 4.2 raises a weird exception. Together, they seem to do the job.

I've set Travis to require the erubi gemfile tests to pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,5 @@ matrix:
   allow_failures:
     - rvm: rbx-3
     - gemfile: test/gemfiles/Gemfile.rails-edge
-    - gemfile: test/gemfiles/Gemfile.rails-5.0.x.erubi
   fast_finish: true
 script: "bundle exec rake submodules test"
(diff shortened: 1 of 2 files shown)


RobinDaugherty:
Looks like there's still an issue with jruby-9.1.12.0. I'm working on it.

RobinDaugherty:
@k0kubun this should be good to merge!

k0kubun:
It's a little complicated, but hopefully this check will be removed after supported Rails becomes higher. I love this well-documented patch. Thank you 👍 👍 👍

amatsuda:
Sorry for not being able to release earlier, but just pubished 5.0.3 gem with this fix.
Thanks! ❤️

Issue #957, opened 2017-09-11
5.0.3 haml -c --stdin generates an exception when checking ruby code


moorecp:
It looks like commit 83da985bf69696362d4f86caa9f5ec74faa58c5d caused an issue when using `--stdin` instead of specifying a filename. If there are no errors in your HAML, when `validate_ruby` is called, `@options[:filename]` is `nil` and an exception is generated. For example:

```
# haml -c --stdin
%foo
  %bar
```

This results in 
```
Exception on line 341: no implicit conversion of nil into String
  Use --trace for backtrace.
```

Previously, this would result in
```
Syntax OK
```

This causes issues with on the fly syntax checkers that utilize STDIN to validate without saving.

k0kubun:
resolved by https://github.com/haml/haml/pull/958

PR #958, opened 2017-09-14
haml -c --stdin fix

TeaMoe:
see #957


k0kubun:
Regression of https://github.com/TeaMoe/haml/commit/83da985bf69696362d4f86caa9f5ec74faa58c5d. Thanks to catch this!

PR #960, opened 2017-10-17
Tidy up --exclude options in .yardopts

erikse:
Updating the `--exclude` path to `plugin.rb`, and removing the entry for `patch.rb` in `.yardopts` file.


k0kubun:
Thanks!

PR #961, opened 2017-10-20
Freeze strings to prevent re-allocation of newline/empty strings

oniofchaos:
Most of these were discovered when profiling performance issues on a page with 1000s of links generated via `link_to` in Rails.

### DIFF
--- a/lib/haml/compiler.rb
+++ b/lib/haml/compiler.rb
@@ -103,7 +103,7 @@ module Haml
       end
 
       if @options[:trace]
-        t[:attributes].merge!({"data-trace" => @options.filename.split('/views').last << ":" << @node.line.to_s})
+        t[:attributes].merge!({"data-trace".freeze => @options.filename.split('/views'.freeze).last << ":".freeze << @node.line.to_s})
       end
 
       push_text("<#{t[:name]}")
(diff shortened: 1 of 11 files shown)


lib/haml/compiler.rb
```diff
@@ -265,6 +265,14 @@ def push_script(text, opts = {})
 
       unless block_given?
         push_generated_script(no_format ? "(#{text}\n).to_s" : build_script_formatter("(#{text}\n)", opts))
+        # TODO this \n should be able to be frozen but the following tests fails
+        #
+        # bin/rails test test/helper_test.rb:596
+        # bin/rails test test/engine_test.rb:534
+        #
+        # Both appear to be testing outer whitespace nuking but at this point
```
oniofchaos: If anyone has insight into this issue and is willing to help me fix the code/test to be able to freeze this string that would be much appreciated! 😍

lib/haml/helpers/action_view_mods.rb
```diff
@@ -55,7 +55,9 @@ module TagHelper
       def content_tag_with_haml(name, *args, &block)
         return content_tag_without_haml(name, *args, &block) unless is_haml?
 
-        preserve = haml_buffer.options.fetch(:preserve, %w[textarea pre code]).include?(name.to_s)
+        @_content_tag_name_cache ||= {}
+        name_string = @_content_tag_name_cache[name] ||= name.to_s
```
oniofchaos: Idea here being that we only needed to run `to_s` on each content tag once. In particular, I was getting `a` in here a lot from links. Any better ideas are welcome.

oniofchaos:
Just noticed the `Haml has two primary branches. stable is where the development of the released version (currently 5.0.4) takes place. This is where most bug fixes should go. Master, on the other hand, is where the next version of Haml is being developed. This is the place for new features.` in the guidelines. Not sure if this is considered a bugfix or not? If so, I can re-make the PR against `stable`.

lib/haml/util.rb
```diff
@@ -207,13 +207,13 @@ def unescape_interpolation(str, escape_html = nil)
...
         else
-          interpolated = if char == '{'
+          interpolated = if char == '{'.freeze
             balance(scan, ?{, ?}, 1)[0][0...-1]
           else
             scan.scan(/\w+/)
           end
-          content = eval('"' + interpolated + '"')
-          content.prepend(char) if char == '@' || char == '$'
+          content = eval("\"#{interpolated}\"")
```
oniofchaos: See https://github.com/haml/haml/pull/961/commits/42f4eaa4cb8339bbefeecfa415169531a6d2c0b3 for justification, summary is that it saves a string allocation each time this code is called for roughly the same performance.

lib/haml/helpers/action_view_mods.rb
oniofchaos: Performance is within margin of error but saves a string allocation for each time this method is called.

```ruby
Warming up --------------------------------------
      master_version   188.121k i/100ms
        fast_version   201.722k i/100ms
      master_version      5.176M (±10.1%) i/s -     25.584M in   5.007528s
        fast_version      5.595M (± 7.3%) i/s -     27.838M in   5.004395s

Comparison:
        fast_version:  5595000.8 i/s
      master_version:  5176423.8 i/s - same-ish: difference falls within error
```

k0kubun: Having `%w[textarea pre code]` as constant would be enough. I guess main performance gain comes from that. I'm reluctant to introduce ivar in template evaluation environment, and in general cache is likely to introduce bug.

k0kubun:
Any performance improvement uglifying code should include benchmark to justify its introduction. But you only attached https://github.com/haml/haml/pull/961#discussion_r146426562.

Basically this change is for compilation, so I want compilation benchmark. Time of compilation + rendering or some rendering benchmark that uses dynamic helper is also welcomed.

lib/haml/compiler.rb
k0kubun: Unfortunately I have no time to assist you, but I would write many `binding.pry`s to see how string that shouldn't be freezed is freezed.

oniofchaos:
I will go ahead and separate the other performance changes unrelated to frozen strings into separate PRs so we can discuss separately and I can more easily post benchmarks. 

Unfortunately all of these files have strings that are mutated in place making the `frozen_string_literal: true` magic comment break tests. See https://github.com/haml/haml/issues/962.  Here's an example from [parser.rb](https://github.com/q-centrix/haml/blob/master/lib/haml/parser.rb#L701). I'm down to assist in rewriting these over time but I don't have the time to do it all at once and I would need some thought partnership in understanding the context of the code in each file.

What do you think about merging the `"str".freeze` changes and then reverting them as we're able to rewrite the files to allow for all strings to be frozen?

k0kubun:
> What do you think about merging the "str".freeze changes and then reverting them as we're able to rewrite the files to allow for all strings to be frozen?

There's no reason to introduce intermediate change for development in master.

Just try adding `.dup` to all string literals and gradually remove them, and make PR after it's done.

lib/haml/helpers/action_view_mods.rb
oniofchaos: @k0kubun I went ahead and added the constant for `%w[textarea pre code] in https://github.com/haml/haml/pull/966. Additionally, what do you think about adding a constant for common `content_tag` options to cache the `name.to_s` call? Mainly the things that show up in [this]( https://github.com/rails/rails/search?l=Ruby&q=content_tag&type=&utf8=%E2%9C%93) search of the Rails repo and any other tag suggestions you have?

k0kubun: And your above benchmark is now expired after you add constant, so you should put benchmark again. 

As I said,

> I'm reluctant to introduce ivar in template evaluation environment, and in general cache is likely to introduce bug.

So at least I wouldn't merge it unless there's SIGNIFICANT performance benefit in benchmark. Basically `to_s` is not slow.

k0kubun:
please rebase against master

k0kubun:
Ah, this PR is split to multiple PRs. Nice! Closing this.

PR #967, opened 2017-10-26
Frozen string literal changes

oniofchaos:
I was able to get `frozen_string_literal: true` turned on in two files. Unfortunately I wasn't able to get them on in the rest, the majority of which were because of various calls to `rstrip!` and `tr!` which are more performant in micro-benchmarks and in memory usage. I made other changes in these files to string builder logic such that they no longer mutate strings, based on [this](https://stackoverflow.com/questions/6549074/whats-the-fastest-way-to-build-a-string-in-ruby) StackOverflow post.

String mutations preventing `frozen_string_literal: true`
`lib/haml/compiler.rb` - [rstrip!](https://github.com/q-centrix/haml/blob/master/lib/haml/compiler.rb#L318)
`lib/haml/filters` - [rstrip!](https://github.com/q-centrix/haml/blob/master/lib/haml/filters.rb#L251)
`lib/haml/template_engine.rb` - [force_encoding](https://github.com/haml/haml/blob/master/lib/haml/temple_engine.rb#L52), [tr!](https://github.com/q-centrix/haml/blob/master/lib/haml/temple_engine.rb#L67)
`lib/haml/parser.rb` - [rstrip!](https://github.com/q-centrix/haml/blob/master/lib/haml/parser.rb#L785)

I wanted to keep the commit history for discussion about the various changes. Once everything is approved I can squash the commits and add a changelog entry.

Result of `ruby benchmark.rb`
```
                                         Haml |     ERB |   Erubi |
-------------------------------------------------------------------
Cached                                  0.088 |   0.073 |   0.051 |
ActionView                             15.558 |         |  10.493 |
ActionView with deep partials          41.512 |         |  37.338 |
```

Current master
```
                                         Haml |     ERB |   Erubi |
-------------------------------------------------------------------
Cached                                  0.085 |   0.073 |   0.070 |
ActionView                             18.001 |         |  10.954 |
ActionView with deep partials          46.022 |         |  34.406 |
```

PR #983, opened 2018-03-01
un-freeze TempleEngine precompiled string literals

### LINKED ISSUE #982: FrozenError - can't modify frozen String
Version: haml `master` branch (no release version yet), tested on commit e4a8be96b5739b2cf06c82fd8be6c43a84b16444.

I produce the following stacktrace when precompiling haml template with `tilt` via a Sinatra application:

```
FrozenError - can't modify frozen String:
	[ruby]/gems/tilt-2.0.8/lib/tilt/template.rb:299:in `force_encoding'
	[ruby]/gems/tilt-2.0.8/lib/tilt/template.rb:299:in `ensure in binary'
	[ruby]/gems/tilt-2.0.8/lib/tilt/template.rb:299:in `binary'
	[ruby]/gems/tilt-2.0.8/lib/tilt/template.rb:289:in `extract_magic_comment'
	[ruby]/gems/tilt-2.0.8/lib/tilt/template.rb:285:in `extract_encoding'
	[ruby]/gems/tilt-2.0.8/lib/tilt/template.rb:191:in `precompiled'
	[ruby]/gems/tilt-2.0.8/lib/tilt/template.rb:251:in `compile_template_method'
	[ruby]/gems/tilt-2.0.8/lib/tilt/template.rb:236:in `block in compiled_method'
	[ruby]/gems/tilt-2.0.8/lib/tilt/template.rb:235:in `synchronize'
	[ruby]/gems/tilt-2.0.8/lib/tilt/template.rb:235:in `compiled_method'
	[ruby]/gems/tilt-2.0.8/lib/tilt/template.rb:169:in `evaluate'
	[ruby]/gems/tilt-2.0.8/lib/tilt/haml.rb:24:in `evaluate'
	[ruby]/gems/tilt-2.0.8/lib/tilt/template.rb:109:in `render'
	[ruby]/gems/sinatra-2.0.1.rc1/lib/sinatra/base.rb:839:in `render'
	[ruby]/gems/sinatra-2.0.1.rc1/lib/sinatra/base.rb:692:in `haml'
```

This issue seems to be caused by setting `frozen_string_literal: true` on `lib/haml/temple_engine.rb` (#967), which breaks an implicit contract in the interface between these two gems.

Whether this is a bug best fixed in `tilt` or `haml` (or both, for best backwards-forwards compatibility) I don't know, but in any case, `tilt`'s `Tilt::HamlTemplate#precompiled_template` expects strings returned from calls to `Haml::TempleEngine` to be not-frozen, because it [calls `#force_encoding` on the returned string](https://github.com/rtomayko/tilt/blob/70000d8af45f7020a88f796b056cc0a1be2410dd/lib/tilt/template.rb#L188-L195).

wjordan:
Fixes #982.
This PR maintains public-API compatibility in `Haml::TempleEngine#precompiled_with_return_value` and `precompiled_with_ambles`, returning mutable strings as expected by `tilt`'s Haml template implementation (which [calls `#force_encoding` on the result](https://github.com/rtomayko/tilt/blob/70000d8af45f7020a88f796b056cc0a1be2410dd/lib/tilt/template.rb#L188-L195)).

### DIFF
--- a/lib/haml/temple_engine.rb
+++ b/lib/haml/temple_engine.rb
@@ -54,7 +54,7 @@ module Haml
     end
 
     def precompiled_with_return_value
-      "#{precompiled};#{precompiled_method_return_value}"
+      "#{precompiled};#{precompiled_method_return_value}".dup
     end
 
     # The source code that is evaluated to produce the Haml document.
@@ -77,7 +77,7 @@ ensure
 @haml_buffer = @haml_buffer.upper if @haml_buffer
 end
 END
-      "#{preamble}#{locals_code(local_names)}#{precompiled}#{postamble}"
+      "#{preamble}#{locals_code(local_names)}#{precompiled}#{postamble}".dup
     end
 
     private


k0kubun:
Ideally mutating `#precompiled` source code would not make to me. But freezing the returned value of public interface is definitely the breaking change, and I agree that it was unwanted side effect of https://github.com/haml/haml/pull/967. Thanks to catch this.

k0kubun:
Probably we should have a test that ensures their returned values can be mutated.

Issue #940, opened 2017-07-06
Javascript escaping invalid, generates SyntaxError: expected expression, got '&'


javinto:
The 5.0.0 release notes refers to: 

> HTML escape interpolated code in filters. #770 (Matt Wildig)
> ```
> :javascript
>     #{JSON.generate(foo: "bar")}
>   Haml 4 output: {"foo":"bar"}
>   Haml 5 output: {&quot;foo&quot;:&quot;bar&quot;}
> ```

However, I have (Haml 5.0.1):

```
:javascript
  var list1= ["a", 'b', 'c'];
  var hash1= {a: "a", b: 'b'};
  var list2= #{JSON.generate( [:a, :b, :c])};
  var hash2= #{JSON.generate(foo: "bar")};
```
In the browser this results into:

```
<script>
  var list1= ["a", 'b', 'c'];
  var hash1= {a: "a", b: 'b'};
  var list2= [&quot;a&quot;,&quot;b&quot;,&quot;c&quot;];
  var hash2= {&quot;foo&quot;:&quot;bar&quot;};
</script>
```

The browsers runs into an error on line 4 (list2) and 5 (hash2): 

> SyntaxError: expected expression, got '&'

k0kubun:
This is intentional, not a bug. You should fix template like this (using data attribute) https://github.com/sferik/rails_admin/pull/2870 or disable HTML escape globally.

javinto:
Okay, a pitty, but I can do this. 

However, the code giving me this problem is like:

:javascript
   var my_var=Array( #{ruby_array}); 

In the HAML Reference file it says: "Currently, filters ignore the :escape_html <http://haml.info/docs/yardoc/Haml/Options.html#escape_html-instance_method> option. This means that #{} interpolation within filters is never HTML-escaped."

Does this mean that the documentation is outdated?

k0kubun:
> Does this mean that the documentation is outdated?

Yes, definitely. I fixed it https://github.com/haml/haml/commit/965e66eb6ba5bbd8b4d89ba8cc60e2a05c658855. Thank you for reporting it.

marz0:
@k0kubun is it possible to set the `escape_html` option for a particular filter? For instance,

```ruby
:javascript{ escape_html: false }
```

Just looking to keep the same behavior as before.

k0kubun:
> Just looking to keep the same behavior as before.

In Rails, you can call html_safe to value. Otherwise no idea.

praxeum:
[...]

PR #984, opened 2018-03-01
add escape_interpolated_html option

wjordan:
Addresses points raised in (closed issue) #940, by allowing optional backwards compatibility to work around breaking changes introduced by #770 (released in [`5.0.0`](https://github.com/haml/haml/blob/master/CHANGELOG.md#500)).

This PR is mostly intended for migrating large/legacy applications written against the v4 API, that want to upgrade to v5 while avoiding time-intensive application refactoring (e.g., https://github.com/haml/haml/issues/940#issuecomment-329769397).

My particular use-case is a need to upgrade `haml` to `>= 5.0.4` for [Ruby 2.5 support](https://github.com/haml/haml/blob/master/CHANGELOG.md#504) without refactoring large amounts of existing application code.

Introduces a new `escape_interpolated_html` option (defaults to the current value of `escape_html` for backwards compatibility with existing `5.x` default behavior), which can be used to override the behavior for escaping HTML in interpolated strings separately from other html-escaping behavior.

With this option, it's possible to set `Haml::Template.options[:escape_interpolated_html] = false` (e.g., in a Rails initializer) for full backwards compatibility with v4 behavior, without any further changes to existing application code.

### DIFF
--- a/lib/haml/temple_engine.rb
+++ b/lib/haml/temple_engine.rb
@@ -13,6 +13,7 @@ module Haml
       encoding:             nil,
       escape_attrs:         true,
       escape_html:          false,
+      escape_interpolated_html: nil,
       filename:             '(haml)',
       format:               :html5,
       hyphenate_data_attrs: true,
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -1121,7 +1121,7 @@ is compiled to
     </div>
 
 Note that `#{}` interpolation within filters is HTML-escaped if you specify
-{Haml::Options#escape_html `:escape_html`} option.
+{Haml::Options#escape_interpolated_html `:escape_interpolated_html`} option.
 
 The functionality of some filters such as Markdown can be provided by many
 different libraries. Usually you don't have to worry about this - you can just
--- a/lib/haml/filters.rb
+++ b/lib/haml/filters.rb
@@ -164,7 +164,11 @@ module Haml
           if contains_interpolation?(text)
             return if options[:suppress_eval]
 
-            text = unescape_interpolation(text, options[:escape_html]).gsub(/(\\+)n/) do |s|
+            escape = options[:escape_interpolated_html]
+            # `escape_interpolated_html` defaults to `escape_html` if unset.
+            escape = options[:escape_html] if escape.nil?
+
+            text = unescape_interpolation(text, escape).gsub(/(\\+)n/) do |s|
               escapes = $1.size
               next s if escapes % 2 == 0
               "#{'\\' * (escapes - 1)}\n"
--- a/lib/haml/options.rb
+++ b/lib/haml/options.rb
@@ -8,7 +8,7 @@ module Haml
     @valid_formats = [:html4, :html5, :xhtml]
 
     @buffer_option_keys = [:autoclose, :preserve, :attr_wrapper, :format,
-      :encoding, :escape_html, :escape_attrs, :hyphenate_data_attrs, :cdata]
+      :encoding, :escape_html, :escape_interpolated_html, :escape_attrs, :hyphenate_data_attrs, :cdata]
 
     # The default option values.
     # @return Hash
@@ -84,6 +84,13 @@ module Haml
     #
     # Defaults to false.
     attr_accessor :escape_html
+
+    # Sets whether or not to escape HTML-sensitive characters in interpolated strings.
+    # See also {file:REFERENCE.md#escaping_html Escaping HTML} and
+    # {file:REFERENCE.md#unescaping_html Unescaping HTML}.
+    #
+    # Defaults to the current value of `escape_html`.
+    attr_accessor :escape_interpolated_html
 
     # The name of the Haml file being parsed.
     # This is only used as information when exceptions are raised. This is


lib/haml/temple_engine.rb
```diff
@@ -13,6 +13,7 @@ class TempleEngine < Temple::Engine
       encoding:             nil,
       escape_attrs:         true,
       escape_html:          false,
+      escape_interpolated_html: nil,
```
k0kubun: To make sure it's a boolean flag, the default value should be `false` instead of `nil`.

k0kubun: Ah, okay I understood you intention. So probably you intends to mean absence of this option here.

k0kubun:
While I understand why you would need this, as it is just a migration path from 4 to 5 and there is no reason to recommend keeping the vulnerable behavior, I don't like to introduce this as official option, which we need to maintain forever. 

I even think having both `escape_attrs` and `escape_html` is crazy. Why do we add the third one? Having this would be convenient for some people, but people who don't know the context would be confused by them.

> My particular use-case is a need to upgrade haml to >= 5.0.4 for Ruby 2.5 support without refactoring large amounts of existing application code.

Given the situation, I would like to have Ruby 2.5 support in https://github.com/haml/haml/tree/stable. Obviously that's the best path to avoid "refactoring large amounts of existing application code", right?

Also, the bug which was fixed for Ruby 2.5 at v5.0.4 was introduced at Haml v5. In my understanding, it doesn't affect Haml v4. Thus I can't think of any reason to upgrade your Haml to 5.0.4+ only for using Ruby 2.5. For your safe Ruby upgrade, please use Haml 4 as it is. If it doesn't work with Ruby 2.5, please report it separately.

k0kubun:
closing due to inactivity

k0kubun:
1. Clarify which filter and what Haml template you're talking about. I can't understand why you mention `.to_json` here at all since filter interpolation could be related to not only JavaScript filter, but also a plain text, CSS, Sass, markdown, ... whatever 
2. Just use `.html_safe` or `raw` if you're talking about Rails. It should be the intended way of embedded dynamic script even in the default template engine Erubi (.erb).
3. Why not use `escape_html: false`?
4. `to_json` is not HTML-safe and completely vulnerable when it's interpolated to the plain text part of HTML. Do not consider only what you care about. The following string is not HTML-escaped properly.

```rb
{a: "<script>alert('hello');</script>"}.to_json #=> "{\"a\":\"<script>alert('hello');</script>\"}"
```

bughit:
rails activesupport version of to_json escapes by default

https://github.com/rails/rails/blob/master/activesupport/lib/active_support/json/encoding.rb#L42

in a rails console your example outputs:
`"{\"a\":\"\\u003cscript\\u003ealert('hello');\\u003c/script\\u003e\"}"`

in the following example (with html encoding turned off) the `.inspect` version injects, the `.to_json` one doesn't.

```haml
- script_injector_string = "</script><script>alert('injected')</script>"

:javascript
  #{script_injector_string.to_json};

:javascript
  #{script_injector_string.inspect};
```

```html
  <body>
    <script>
  "\u003c/script\u003e\u003cscript\u003ealert('injected')\u003c/script\u003e";
</script>
<script>
  "</script><script>alert('injected')</script>";
</script>
  </body>
```

> Why not use escape_html: false

To avoid porting old code on upgrade.

bughit:
>>Why not use escape_html: false

>To avoid porting old code on upgrade.

This was intended for the html_safe/raw suggestion

bughit:
> Just use .html_safe

if to_json did not escape, calling html_safe/raw on it is not good advice, because you end up where you started, with a theoretical possibility of script injection.  The following injects.

```haml
- script_injector_string = "</script><script>alert('injected')</script>"

:javascript
  #{JSON.dump(script_injector_string).html_safe};
```
```html

If you did not want to depend on the protection of rails to_json, the proper workaround would be to let haml html escape the to_json output, wrap it in quotes and then on the javascript side,  html decode it, then JSON.parse the result. Considering that html decoding is not even a js builtin, this is getting pretty cumbersome.

I hope this underscores the need for this option.  If you are already using to_json, having to port to .to_json.html_safe is entirely pointless, as you are still protected by to_json.  Much better to set an option and leave the already safe and simple code alone.

bughit:
@k0kubun so would you consider adding an `escape_filter_interpolations` option, in light of

https://github.com/haml/haml/pull/984#issuecomment-455755480

and

https://github.com/haml/haml/pull/984#issuecomment-455758957

k0kubun:
> rails activesupport version of to_json escapes by default
> https://github.com/rails/rails/blob/master/activesupport/lib/active_support/json/encoding.rb#L42
> in a rails console your example outputs:
> "{\"a\":\"\\u003cscript\\u003ealert('hello');\\u003c/script\\u003e\"}"

Interesting. So this is really Rails-specific issue. Ideally this discussion should go to haml-rails repository but Haml's extension for Rails still exists in haml for now, so I'll talk about it here for now.

Could you confirm the following parts?

1. In a `:javascript` filter, you want to embed Rails-specific-`.to_json`-ed value which is NOT `.html_safe`-ed even while it's actually HTML-safe by the `.to_json`'s special escape.
2. Your final goal is to pass a JSON-compatible value to JavaScript script in `:javascript` filter. Why don't you use data attribute to do that, which is known to be the safest way?

> To avoid porting old code on upgrade.

The affected templates are only Haml templates with `:javascript` filter embedding `.to_json`-values, right? Do you have so many such templates?

Ultimately, I would recommend using data attribute to pass values to JavaScript safely, so I don't think this is a reasonable use case and I'm reluctant to support such feature to let people move towards that way.

<hr>

I hope gives the answers to both comments. I'm actually not recommending `raw` / `.html_safe` but suggesting to use safer way to pass values to JavaScript regardless of the legacy Haml 4 feature you want to resurrect. In that sense, I reject introducing `escape_interpolated_html` for your use case. Please provide another one if you need.

bughit:
> I reject introducing escape_interpolated_html for your use case. Please provide another one if you need.

Small aside, escape_interpolated_html is not a good name, it reads the same as the exiting escape_html, the new one would need to be called escape_filter_interpolations or just escape_filters.

I am not clear on what you are saying I could provide?  Another use-case?

bughit:
thanks, confirming that it's working

PR #995, opened 2018-10-29
add verbocity

eitangenis:


k0kubun:
Too few descriptions and tests to accept. You can achieve this without having this in master by creating a wrapper of `Haml::Engine` or overriding `Haml::Engine#initialize` by prepending module. That should suffice for your own debugging. You should describe why explain must be committed to this repository.

In addition to that, having `"\n"` before `template` is completely illegal in template engine's world because it breaks the consistency with the line number of backtrace and the equivalent in the original template. So this is hard to merge as is.

k0kubun:
You're submitting PRs to your own fork, so I guess this time you mistakenly submitted it to this upstream. If so, never mind about my above comment and enjoy updating your fork.

Issue #990, opened 2018-08-26
escape_html option being ignored when trying to escape html?


ckuttruff:
haml version: `5.0.4`  (latest at the time of this issue filing)
env: sinatra (`2.0.3`)

hey all, thanks for all your work.  I love haml.  It's very nice.

One thing I ran across when I grew tired of subbing all output directives to `&=`...
When I set the following in my toplevel sinatra controller:
`Haml::Options.defaults[:escape_html] = true`
(according to docs: http://haml.info/docs/yardoc/file.REFERENCE.html#escaping_html)

I don't see `=` having the same effect as `&=`.  `&=` properly escapes while `=` does not.

Am I missing something basic here or can y'all reproduce?
Please let me know if I can provide any more information or context.  Thanks!

ckuttruff:
btw, I have verified with inspect statements that the option is indeed set.  But again `=` is still rendering html markup while `&=` is not.

TIA

k0kubun:
Did you try `Haml::Template.options[:escape_html] = true`?

ckuttruff:
Sorry for late reply.  I'll def give rhis a shot and report back.  Setting reminder.
Thanks

ckuttruff:
@k0kubun Where would that config go btw?  Sorry, I'm running a simple sinatra app and have tried sticking it in a few obvious places but keep getting:
`uninitialized constant Haml::Template`

k0kubun:
Oh yeah, I remembered Tilt has its own `Tilt::HamlTemplate`... both wouldn't work. Maybe `set :haml, { escape_html: false }` on Sinatra?

k0kubun:
It's likely to work. Could you try `set :haml, { escape_html: true }`?

ckuttruff:
lol, perfect, I have an html soup of refusing to render html.

Thank you!  Would you like me to submit a PR on the docs site or somewhere to specifically note that?  Would love to help out future sinatra/haml users there!

Thanks for quick response and resolution.

k0kubun:
PR that fixes somewhere of https://github.com/haml/haml/blob/master/REFERENCE.md would be appreciated.

ckuttruff:
will do that now.

k0kubun:
I don't have commit access of that repository and I don't know how it's maintained (it should have README)... but perhaps it's auto-generated from https://github.com/haml/haml/blob/master/REFERENCE.md according to its contents, and thus you should have sent PR for that as I said. Anyway thanks.

[...]

PR #996, opened 2018-11-01
add note about sinatra for setting config options

ckuttruff:


k0kubun:
remembered https://github.com/haml/haml/issues/990 now. Thank you!

PR #1002, opened 2018-12-03
Require Util in Engine, because Engine includes Util

jcoyne:


k0kubun:
then, why not in Parser? why not in Compiler? It's shared globally and I couldn't imagine how this would improve the maintainability without any description. Closing since I personally prefer the current code.

jcoyne:
I was doing this because I wanted to use haml, but not the railtie.  It would be more convenient to do:
```
 require 'haml/engine'
```

rather than
```
require 'haml/util'
require 'haml/engine'
```

k0kubun:
Splitting Rails extensions from haml.gem to haml-rails.gem was planned 2 years ago and that should be the solution for your use case. Otherwise a change like yours could be easily reverted unexpectedly. 

Unfortunately it requires a lot of work of release engineering of both haml.gem and haml-rails.gem, so it's not done yet. Please wait until the gems' owners (not including me) have time for it.

PR #1008, opened 2019-03-15
Add support to `Haml::Plugin` for Rails 6

sebjacobs:
In Rails 6 the API for template handlers is changing, a template handler
must now take two arguments [1], the template and the source, otherwise you
will see the following deprecation warning:

```
ActiveSupport::DeprecationException: DEPRECATION WARNING: Single arity
template handlers are deprecated.  Template handlers must now accept two
parameters, the view object and the source for the view object.
Change:
  >> Class#call(template)
To:
  >> Class#call(template, source)
```

I have attempted to retain support for older versions of Rails by
providing a default source of `nil`.

[1] https://www.github.com/rails/rails/commit/28f88e0074

### DIFF
--- a/lib/haml/plugin.rb
+++ b/lib/haml/plugin.rb
@@ -5,7 +5,7 @@ module Haml
   class Plugin
     def handles_encoding?; true; end
 
-    def compile(template)
+    def compile(template, source)
       options = Haml::Template.options.dup
       if template.respond_to?(:type)
         options[:mime_type] = template.type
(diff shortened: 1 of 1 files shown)


amatsuda:
Yeah, I know this would be the workaround for the currently released Rails 6 beta gems, but I'm not sure whether Aaron's "refactoring" has been all done or still ongoing...

So for now, let's just merge this and wait until another stabler Rails release.

@sebjacobs Thank you for the patch anyway!

morgoth:
@amatsuda Rails 6 is around the corner ;-) - can we have this released?

Issue #1014, opened 2019-06-12
.text.haml files have html escaped content


cesarizu:
On haml 5.1.1, all text in `.text.haml` files is html escaped. Using `escape_filter_interpolations` doesn't seem to help. This seems to be related: https://github.com/rails/rails/pull/8235

Is there any way to disable escaping on `text.haml` files altogether?

cesarizu:
PR at haml/haml#1015

hcatlin:
Merged in the PR!

PR #1015, opened 2019-07-17
Do not run escape_html on text/plain files

cesarizu:
Currently `text.haml` files are html escaped. This PR fixes that.

Related to haml/haml#1014 and haml/haml-rails#163

### DIFF
--- a/lib/haml/parser.rb
+++ b/lib/haml/parser.rb
@@ -307,7 +307,7 @@ module Haml
         return ParseNode.new(:plain, line.index + 1, :text => line.text)
       end
 
-      escape_html = @options.escape_html if escape_html.nil?
+      escape_html = @options.escape_html && @options.mime_type != 'text/plain' if escape_html.nil?
       line.text = unescape_interpolation(line.text, escape_html)
       script(line, false)
     end
--- a/test/engine_test.rb
+++ b/test/engine_test.rb
@@ -1458,6 +1458,10 @@ HAML
     end
   end
 
+  def test_mime_type_text_plain_with_interpolation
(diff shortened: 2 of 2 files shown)


lib/haml/compiler.rb
```diff
@@ -277,7 +277,7 @@ def push_script(text, opts = {})
 
     def build_script_formatter(text, opts)
       text = "(#{text}).to_s"
-      if opts[:escape_html]
+      if opts[:escape_html] && !opts[:mime_type] == 'text/plain'
```
jerryskye: I think it would be more clear to use the inequality operator here :blush:
```suggestion
      if opts[:escape_html] && opts[:mime_type] != 'text/plain'
```

cesarizu: Thanks for spotting that. I'm not really sure why I wrote it that way. I've updated the PR.

cecilian:
Hey, thanks for this PR! @mantainers, will it be merged anytime soon?

hcatlin:
Sorry it's been a ghost town around here– but we'd merge this in.... if you add in tests!

cesarizu:
I've updated the commit with a test. It seems weird to me that `@options.escape_html` is true when `@options.mime_type == 'text/plain'` (at least on rails). Maybe the fix needs to go somewhere else?

hcatlin:
I'm not 100% sure why you'd use Haml to make a text file... but heck, good for you!

cesarizu:
Thanks for merging this, we have some use cases for plain text using haml, namely generating csv files, text emails, text versions of some pages that use some partials that can be called from either html templates or text templates.

PR #1016, opened 2019-08-06
Unfreeze string literals for ParseNode#inspect

jalyna:
Since newrelic is using the `#inspect` method on `ParseNode` the ruby code crashes due to frozen strings in `inspect`, which are recursively called with `gsub!` causing this error:

Screen Shot 2019-08-06 at 10 54 08

I added a regression test that fails with the current code.

### DIFF
--- a/lib/haml/parser.rb
+++ b/lib/haml/parser.rb
@@ -202,7 +202,7 @@ module Haml
       end
 
       def inspect
-        %Q[(#{type} #{value.inspect}#{children.each_with_object('') {|c, s| s << "\n#{c.inspect.gsub!(/^/, '  ')}"}})]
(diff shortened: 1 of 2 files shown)


amatsuda:
@jalyna Thank you for the patch!
I understand that we need the first `children.each_with_object(''.dup)` modification, but I don't see why we need the second `.dup` call at the end of that line.

jalyna:
@amatsuda this is a really good question! the second `.dup` is still needed because `%Q[]` creates a frozen string literal (still). The test will fail if it gets removed.

```rb
# frozen_string_literal: true

puts %Q[hello world].frozen?
```

returns true.

amatsuda:
@jalyna Ah! I just got what you meant by "which are recursively called". Thank you for a clear explanation (and I'm sorry that I wasn't carefully reading it...)!

amatsuda:
@jalyna Just published version 5.1.2 gem with this patch. Thanks!

PR #1020, opened 2019-12-04
Attributes wrapped correctly on escape_attrs=false

faelsoto:
When an attribute has quotes, and we need to disable `escape_attrs`, the quotes aren't escaped correctly, this PR fixes it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ Note that you can also run just one test out of the test suite if you're working
 on a specific area:
 
 ~~~sh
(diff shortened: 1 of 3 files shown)


hcatlin:
Seems like #1028 covers this case and more... let us know if that doesn't cover it!

faelsoto:
Sorry I took this long to answer, I just had a few minutes to check it, and it seems that when `escape_attrs` is set to false, the HTML gets messy still.

image

PR #1028, opened 2020-06-22
Escape attributes regardless of whether it's SafeBuffer or not

### LINKED ISSUE #993: attribute with safe-buffer containing a apostrophe is not escaped
I've run into an issue when embedding html in an attribute in haml. 

if the attribute is a `SafeBuffer` (or otherwise `html_safe?` already) then any apostrophes don't get converted to `&#39`. I thought it should -- because although it is marked html_safe, that doesn't necessarily mark it as attribute-safe (hence the need for `'` to `&#39` conversion).

Is this something Haml should support? I've got a test for it (but no clue how to implement as yet). 

The alternative is that this change has to be done within the template itself (with something like `my_safe_buffer.gsub("'", "&#39;").html_safe`)

### LINKED ISSUE #1019: escape_attrs wraps incorrectly?
Regarding the same idea of #984, we're migrating an app to newer Ruby/Rails versions, and it seems that the 4.x behavior isn't complete with just `escape_interpolated_html`, as the attributes are being a little weird.

On 4.x:

```ruby
Haml::Engine.new(%{ %a{ :href => '/', :'@click' => "callback('a')" } link }.strip).render
=> <a @click="callback('a')" href='/'>link</a>
```

On 5.1.2:

```ruby
Haml::Engine.new(%{ %a{ :href => '/', :'@click' => "callback('a')" } link }.strip).render
=> <a @click='callback(&#39;a&#39;)' href='/'>link</a>
```

This change breaks the inline templates for libraries like Vue. Digging in the code, I find `escape_attrs`, which doesn't help on either version, both outputting:

```ruby
Haml::Engine.new(%{ %a{ :href => '/', :'@click' => "callback('a')" } link }.strip, :escape_attrs => false ).render
=> <a @click='callback('a')' href='/'>link</a>
```

Is this the expected output? I'm not seeing a case in which I would like to break out of the attribute value, I think this is a security issue.

I found that the `attr_wrapper` option can be assigned to a symbol that will wrap the attribute (single quotes by default), but this would break whenever the attribute value has the same value.

My proposal is to call inspect for the attribute value:

```
"callback('a', \"b\")".inspect
=> "callback('a', \"b\")"
```

Does that make sense? Should I start working on a PR or is it part of the spec or wontfix?

k0kubun:
Fixes #993, fixes #1019

Like rails helpers fixed in CVE-2016-6316, an attribute should be HTML-escaped even if it's a SafeBuffer.

### DIFF
--- a/lib/haml/escapable.rb
+++ b/lib/haml/escapable.rb
@@ -4,30 +4,31 @@ module Haml
   # Like Temple::Filters::Escapable, but with support for escaping by
   # Haml::Herlpers.html_escape and Haml::Herlpers.escape_once.
   class Escapable < Temple::Filter
+    # Special value of `flag` to ignore html_safe?
+    EscapeSafeBuffer = Struct.new(:value)
+
     def initialize(*)
       super
-      @escape_code = "::Haml::Helpers.html_escape((%s))"
-      @escaper = eval("proc {|v| #{@escape_code % 'v'} }")
-      @once_escape_code = "::Haml::Helpers.escape_once((%s))"
-      @once_escaper = eval("proc {|v| #{@once_escape_code % 'v'} }")
       @escape = false
+      @escape_safe_buffer = false
     end
 
     def on_escape(flag, exp)
-      old = @escape
-      @escape = flag
+      old_escape, old_escape_safe_buffer = @escape, @escape_safe_buffer
+      @escape_safe_buffer = flag.is_a?(EscapeSafeBuffer)
+      @escape = @escape_safe_buffer ? flag.value : flag
       compile(exp)
     ensure
-      @escape = old
+      @escape, @escape_safe_buffer = old_escape, old_escape_safe_buffer
     end
 
     # The same as Haml::AttributeBuilder.build_attributes
     def on_static(value)
       [:static,
        if @escape == :once
-         @once_escaper[value]
+         escape_once(value)
        elsif @escape
-         @escaper[value]
+         escape(value)
        else
          value
        end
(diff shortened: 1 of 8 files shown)


lib/haml/helpers.rb
```diff
@@ -607,9 +607,12 @@ def haml_tag_if(condition, *tag)
     # @param text [String] The string to sanitize
     # @return [String] The sanitized string
     def html_escape(text)
-      ERB::Util.html_escape(text)
```
k0kubun, line 610: ActiveSupport monkey-patches this method for `html_safe?`

lib/haml/escapable.rb
```diff
@@ -4,30 +4,31 @@ module Haml
   # Like Temple::Filters::Escapable, but with support for escaping by
   # Haml::Herlpers.html_escape and Haml::Herlpers.escape_once.
   class Escapable < Temple::Filter
```
hcatlin, line 6: Why do we need our own escapable implementation?

lib/haml/helpers.rb
hcatlin: Definitely like seeing the `CGI` version vs leaning on ActiveSupport

test/template_test.rb
```diff
@@ -248,7 +248,18 @@ def test_xss_protection_in_attributes
...
+<meta content='&#39;&quot;' />
+<meta content='&#39;&quot;' />
+<meta content='&#39;&quot;' />
+HTML
+%meta{ content: %{'"}.html_safe }
+- val = %{'"}.html_safe
+%meta{ content: val }
+- hash = { content: val }
+%meta{ hash }
+HAML
```
hcatlin, line 262: Can you pull this out into actual template files, instead of having this pasted into this already-large file? Should be as easy as creating some files in the `templates` and `results` folder.

I'd really like to get rid of these assertion-based tests in this test suite that is focused on full-template file comparison.

lib/haml/escapable.rb
k0kubun: That's because Haml has its original way of escaping: `escape_once`. We can't share such logic with any other libraries.

test/template_test.rb
k0kubun: done https://github.com/haml/haml/pull/1028/commits/da0e7da5c89a54f067427ed7ece5e15eb3e5a178

hcatlin:
I think you misunderstood what I meant about the files... let me show you in a next PR.

gurgeous:
Pardon my ignorance here - won't this double escape?

```ruby
# in rails console
Haml::Engine.new("%p{title: html_escape('hi &')}").render
=> "<p title='hi &amp;amp;'></p>\n"
```

That seems incorrect and is causing problems for my app. Should we adopt the Rails fix and only escape the quotes? See https://github.com/rails/rails/compare/v4.2.7...v4.2.7.1

``` ruby
        def tag_option(key, value, escape)
          if value.is_a?(Array)
            value = escape ? safe_join(value, " ") : value.join(" ")
          else
            value = escape ? ERB::Util.unwrapped_html_escape(value) : value
          end
          %(#{key}="#{value.gsub(/"/, '&quot;'.freeze)}")
        end
```

k0kubun:
Could you open a separate issue or pull request? Because that is an intended behavior in this PR, it seems like a new feature. Double escaping could happen for many other cases when you escape something where escaping automatically happens, and that place is where you don't need to escape it since Haml 5+.

gurgeous:
Yes, of course, thank you for the quick response!

PR #1029, opened 2020-06-23
Fix Test Suite with Bundler

hcatlin:
The README specifies that you can run `rake` in order to successfully pass the test suite.

Unfortunately, due to the include order in the test_helper, you actually have to run `bundler exec rake`.

We were already partially setting up Bundler in the `test_helper`, and this makes it startup properly.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -14,8 +14,7 @@ if ENV["COVERAGE"]
   SimpleCov.start
 end
 
-require 'bundler'
-Bundler.setup
+require 'bundler/setup'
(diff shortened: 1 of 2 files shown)


test/test_helper.rb
```diff
@@ -15,6 +15,7 @@
 end
 
 require 'bundler'
```
k0kubun, line 17:
```suggestion
require 'bundler/setup'
```
can do the same thing

hcatlin: Oh, thanks for the hint.

PR #1033, opened 2020-07-14
Reorder Class Names Follow Up

hcatlin:
Update the haml-spec submodule, added text to the REFERENCE file specifying this behaviour, and also added a more comprehensive test case.


hcatlin:
Aggro-merging since master build is broken without the new spec

PR #1034, opened 2020-07-15
Breaking Up the Monolithic Engine Tests

hcatlin:
That was a lot of code. Phew.

Broke up the giant file into a 22 different test suites.

This is why we can't have nice things!


hcatlin:
Ugh, the tests are broken in versions of Ruby before 2.3. That's because I started to implement the squiggle heredoc https://infinum.com/the-capsized-eight/multiline-strings-ruby-2-3-0-the-squiggly-heredoc

These massively improve the readability of our tests... and I think are well worth it.

So– Ruby 2.2 has been EOL since 2018-03-31 according to– https://www.ruby-lang.org/en/downloads/branches/

I'm cleaning up our test matrix to only test against non-EOL Ruby versions.

If you aren't updating your Ruby version, then you sure as hell aren't updating your Haml-gem.

Our apps will continue to work– but we won't test against it.

k0kubun:
I got my baby yesterday and I'm extremely busy. Please don't ask a review to me.

hcatlin:
Congratulations! Yes, go be a good parent!

PR #1039, opened 2020-09-10
Enable frozen string literal

aliismayilov:
https://github.com/haml/haml/pull/967 added the frozen string literal magic comments on all ruby files.

To ensure that the strings are not modified in the future, I think the tests also should cover it. This PR fixes a string mutation case and runs tests with `RUBYOPT=--enable-frozen-string-literal`, where the ruby interpreter will ensure the frozen strings.

### DIFF
--- a/lib/haml/util.rb
+++ b/lib/haml/util.rb
@@ -213,7 +213,7 @@ MSG
             scan.scan(/\w+/)
           end
           content = eval("\"#{interpolated}\"")
-          content.prepend(char) if char == '@' || char == '$'
+          content = "#{char}#{content}" if char == '@' || char == '$'
           content = "Haml::Helpers.html_escape((#{content}))" if escape_html
 
           res << "\#{#{content}}"
--- a/test/cases/engine_internals_test.rb
+++ b/test/cases/engine_internals_test.rb
@@ -18,6 +18,10 @@ class EngineInternalsTest < TestBase
     Haml::Options.defaults.replace(defaults)
   end
 
+  # This test is here due to https://github.com/haml/haml/issues/972
+  # where we used to trust in the #inspect of `true` and `false` to return
+  # "true" and "false". This tests isolates and executes a whole ruby interpreter
+  # to catch the problem.
   def test_engine_inspect_monkeypatch
     Tempfile.open(['haml-test', '.rb']) do |f|
       f.puts <<-HAML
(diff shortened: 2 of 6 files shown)


aliismayilov:
I'm not sure how to deal with jruby warning causing the tests to fail. Any input would be welcome

hcatlin:
Ooff, that's annoying.

lib/haml/util.rb
```diff
@@ -213,7 +213,7 @@ def unescape_interpolation(str, escape_html = nil)
             scan.scan(/\w+/)
           end
           content = eval("\"#{interpolated}\"")
-          content.prepend(char) if char == '@' || char == '$'
+          content = "#{char}#{content}" if char == '@' || char == '$'
```
hcatlin, line 216: I'd be curious if there are any performance issues here...?

test/cases/engine_internals_test.rb
```diff
@@ -37,7 +37,7 @@ def inspect
         print Haml::Engine.new('%div{ foo: true, bar: false }').render
       HAML
       f.close
-      out = IO.popen([RbConfig.ruby, f.path], &:read)
+      out = IO.popen([RbConfig.ruby, '-W0', f.path], &:read)
```
hcatlin, line 40: Can you add a documentation note here on why the `-W0` is there and what it's used for?

test/cases/nesting_test.rb
```diff
@@ -160,7 +160,7 @@ def test_silent_end_with_stuff
 b
 a
 HTML
-- str = "abcde"
+- str = +"abcde"
```
hcatlin, line 163: Wow, I had no idea about this feature... it took me some serious digging to learn what it even was. Great work!

hcatlin:
Ran the tests locally with the same version of JRuby and it passed just fine... still investiating.

hcatlin:
Since the intent of the breaking test is to capture an interesting edge case, I have slightly widened it's expectations, so that we don't have to be 100% sure that ALL interpreters NEVER print a warning when executed as a child or in various different configurations.

hcatlin:
Further, this fails on the `main` branch: https://travis-ci.org/github/haml/haml/jobs/731023617

test/cases/engine_internals_test.rb
aliismayilov: here the I attempted to set ruby interpreter warning level to 0 (=silence), with hopes that jruby warning wouldn't be printed out. I see that you've already merged the PR, I guess, I can leave it as is.

PR #1043, opened 2020-10-20
Implement multiline attributes

### LINKED ISSUE #981: Proposal: Multiline Attributes
Following off of the discussion in #976....

As I read back on those old comments and in the documentation... we only were thinking of the wall for content text blobs to stop people from writing too much 'pure ruby' in weird places. That's why we made it hard. Actually, we almost were explicitly thinking of things after the `=`

However... stylistically, I think we ended up punishing interesting (and pretty!) Haml-styling possibilities. Consider this.

```
%div{
  myAttribute: "blue",
  anotherAttribute: "red",
  moreStuff: "My text is #{@var}",
}
  I am some nested content.
```

This doesn't seem ugly to me. That's not something we really considered at the time, as we were just trying to keep logic out of the templates, but, I don't think we were thinking of a React-y world where it was a good idea to have multiple attributes on an element.

Maybe we should be more lax (though, not with the indentation!) about what happens between the `{` and the `}`?

k0kubun:
Fixes #981.

This is ported from [Hamlit v1 code](https://github.com/k0kubun/hamlit/blob/v1.7.2/lib/hamlit/parsers/attribute.rb#L29-L43) which supported multiline attributes. It was dropped when Hamlit started using Haml's parser in v2, but I'd be happy to resurrect the support. Please read the code comments for design decision details.

### DIFF
--- a/lib/haml/parser.rb
+++ b/lib/haml/parser.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'ripper'
 require 'strscan'
 
 module Haml
@@ -89,6 +90,9 @@ module Haml
 
     ID_KEY    = 'id'.freeze
     CLASS_KEY = 'class'.freeze
+
+    # Used for scanning old attributes, substituting the first '{'
+    METHOD_CALL_PREFIX = 'a('
 
     def initialize(options)
       @options = Options.wrap(options)
@@ -651,13 +655,18 @@ module Haml
     # @return [String] rest
     # @return [Integer] last_line
     def parse_old_attributes(text)
-      text = text.dup
       last_line = @line.index + 1
 
       begin
-        attributes_hash, rest = balance(text, ?{, ?})
+        # Old attributes often look like a valid Hash literal, but it sometimes allow code like
+        # `{ hash, foo: bar }`, which is compiled to `_hamlout.attributes({}, nil, hash, foo: bar)`.
+        #
+        # To scan such code correctly, this scans `a( hash, foo: bar }` instead, stops when there is
+        # 1 more :on_embexpr_end (the last '}') than :on_embexpr_beg, and resurrects '{' afterwards.
+        balanced, rest = balance_tokens(text.sub(?{, METHOD_CALL_PREFIX), :on_embexpr_beg, :on_embexpr_end, count: 1)
+        attributes_hash = balanced.sub(METHOD_CALL_PREFIX, ?{)
       rescue SyntaxError => e
-        if text.strip[-1] == ?, && e.message == Error.message(:unbalanced_brackets)
+        if e.message == Error.message(:unbalanced_brackets) && !@template.empty?
           text << "\n#{@next_line.text}"
           last_line += 1
           next_line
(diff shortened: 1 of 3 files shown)


hcatlin:
Whoa... I didn't see this! @k0kubun– you... you.... <3

kshnurov:
A simple `%td{ title: 'asd' } asd'qwe` fails with `Old attributes must start with "{" and end with "}"` because of that extra ' in the content. 5.2.0 works fine.
https://github.com/haml/haml/issues/1062

k0kubun:
Sorry, I couldn't reproduce it. Could you prepare a repository that reproduces your issue? Thanks.

```
$ cat /tmp/a.haml
%td{ title: 'asd' } asd'qwe

$ ruby -v 
ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-darwin19]

$ haml -v
Haml 5.2.1

$ haml /tmp/a.haml
<td title='asd'>asd'qwe</td>
```

kshnurov:
@k0kubun turns out it's a Ruby 2.4 issue. 2.5, 2.6, and 2.7 work fine.
```
$ cat ./a.html
%td{ title: 'asd' } asd'qwe

$ ruby -v
ruby 2.4.10p364 (2020-03-31 revision 67879) [x86_64-darwin18]

$ haml -v
Haml 5.2.1

$ haml ./a.html
Exception on line 219: Old attributes must start with "{" and end with "}"
  Use --trace for backtrace.
```

k0kubun:
Yeah, it seems like a bug of Ripper. The problem is that the parser reorders tokens in edge cases like that, and it's a bit hard to implement a fix on Haml's side. Thus the bug report should go to ruby-core, instead of this haml repository.

```
$ ruby -v -rripper -e "puts Ripper.lex(%q[a( b: 'c' } d'e]).map{|x| x[2]}.join"
ruby 2.4.9p362 (2019-10-02 revision 67824) [x86_64-darwin18]
ea( b: 'c' } d'

$  ruby -v -rripper -e "puts Ripper.lex(%q[a( b: 'c' } d'e]).map{|x| x[2]}.join"
ruby 2.5.8p224 (2020-03-31 revision 67882) [x86_64-darwin19]
a( b: 'c' } d'e
```

JFYI, 2.3 also works fine. 2.2 and older have the bug, like 2.4.

```
$ ruby -v -rripper -e "puts Ripper.lex(%q[a( b: 'c' } d'e]).map{|x| x[2]}.join"
ruby 2.3.8p459 (2018-10-18 revision 65136) [x86_64-darwin18]
a( b: 'c' } d'e

kshnurov:
@k0kubun 2.4 is EOL so there won't be a fix. Maybe it's time for HAML to drop Ruby 2.4- compatibility? @hcatlin

hcatlin:
@kshnurov what would that mean to you? We only test against Ruby 2.7+, and I wouldn't think we should error unless necessary.

kshnurov:
@hcatlin README says `Haml currently supports Ruby 2.0.0 and higher`. I guess Ruby 2.4- users shouldn't be confused and upgrade to the latest versions of Haml.

k0kubun:
> We only test against Ruby 2.7+

Sorry for confusing you maybe by leaving a premature GitHub Actions config, but we currently test 2.5+ in Travis https://github.com/haml/haml/blob/d3839d319af10f81aebf121cb97f3b11b3dd9e94/.travis.yml#L9-L10

If we want to clarify the supported versions, https://github.com/haml/haml/blob/d3839d319af10f81aebf121cb97f3b11b3dd9e94/haml.gemspec#L26 would be the best place to do so. (Of course we should update README as well like @kshnurov said)

hcatlin:
Thanks for the clarification everyone.

PR #1050, opened 2021-02-17
Add support for annotate_rendered_view_with_filenames

jdelStrother:
Rails 6.1 adds a `ActionView::Base.annotate_rendered_view_with_filenames` option, but it's ERB-only (https://github.com/rails/rails/blob/89b52a8dd35cd90e0de05aa2ddc3f5b61832a269/actionview/lib/action_view/template/handlers/erb.rb#L53-L56)

How about something like this?  It adds filename comments into the html looking like, eg:

```html
<div class='navbar-light navbar-collapse collapse wi-100 d-md-none' id='navbar-collapse'>
    <div class='navbar-nav phm mbm' role='menu'>
        <!-- BEGIN app/views/page_header/_user_profile_links.html.haml -->
        <a class="nav-link" href="/profile">Profile</a>
        <a class="nav-link" href="/account/edit">Settings</a>
        <div class='dropdown-divider'></div>
        <a class="nav-link" href="/logout">Log out</a>
        <!-- END app/views/page_header/_user_profile_links.html.haml -->

```

Not sure how clean the implementation is, let me know if there's a more suitable place to insert these.

### DIFF
--- a/lib/haml/plugin.rb
+++ b/lib/haml/plugin.rb
@@ -4,6 +4,11 @@ module Haml
 
   # This module makes Haml work with Rails using the template handler API.
   class Plugin
+    class << self
+      attr_accessor :annotate_rendered_view_with_filenames
+    end
+    self.annotate_rendered_view_with_filenames = false
+
     def handles_encoding?; true; end
 
     def compile(template, source)
@@ -14,9 +19,19 @@ module Haml
         options[:mime_type] = template.mime_type
       end
       options[:filename] = template.identifier
+
+      preamble = '@output_buffer = output_buffer ||= ActionView::OutputBuffer.new if defined?(ActionView::OutputBuffer);'
+      postamble = ''
+
+      if self.class.annotate_rendered_view_with_filenames
+        preamble += "haml_concat '<!-- BEGIN #{template.short_identifier} -->'.html_safe;"
+        postamble += "haml_concat '<!-- END #{template.short_identifier} -->'.html_safe;"
+      end
+
       Haml::Engine.new(source, options).compiler.precompiled_with_ambles(
         [],
-        after_preamble: '@output_buffer = output_buffer ||= ActionView::OutputBuffer.new if defined?(ActionView::OutputBuffer)',
+        after_preamble: preamble,
+        before_postamble: postamble
       )
     end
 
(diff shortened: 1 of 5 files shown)


jdelStrother:
Pushed a fix for Rails < 6. I think the remaining test failures are a recently introduced problem with jruby (https://github.com/jruby/jruby/issues/6547)

lib/haml/plugin.rb
```diff
@@ -4,6 +4,11 @@ module Haml
 
   # This module makes Haml work with Rails using the template handler API.
   class Plugin
+    class << self
+      attr_accessor :annotate_rendered_view_with_filenames
+    end
+    self.annotate_rendered_view_with_filenames = false
```
k0kubun, line 10: Instead of creating Haml's own config, how about checking `ActionView::Base.try(:annotate_rendered_view_with_filenames)`? I think it's what Rails checks for .erb files too, and I don't think we need to support it with Rails versions that don't support it for .erb files.

k0kubun: Oh sorry I overlooked what you wrote in `lib/haml/railtie.rb`. But at least I want this `Haml::Plugin.annotate_rendered_view_with_filenames` to either be private or not exist for maintainability of Haml and consistency with .erb files.

jdelStrother: Could do - I did it this way because it seemed like it would be useful even outside of Rails. But if it'll help get this merged, I'll update it.

dmarcoux:
Could I help somehow in getting this merged?

lib/haml/plugin.rb
hcatlin: I think I agree with @jdelStrother here... as really Haml::Plugin is for ActionView, and the railtie configures it with the Rails App settings. I mean, at this point the whole abstraction of the Plugin is kind of weird anyway and probably needs revisiting in later versions.

The main advantage of doing this as it's 'own setting' is it's just much more testable too.

jdelStrother:
@hcatlin thanks!

hcatlin:
@jdelStrother @dmarcoux https://github.com/haml/haml/releases/tag/v5.2.2

dmarcoux:
Thank you @hcatlin!

PR #1055, opened 2021-06-06
Drop Haml helpers

k0kubun:
I was asked by @hcatlin to bring Hamlit's speed to Haml, breaking backward compatibility as needed. This PR demonstrates what it _at least_ takes. You cannot use `= preserve` or `= find_and_preserve`, but still `~` works and it internally calls `Haml::Helpers.find_and_preserve`.

We also need to get rid of `Haml::Helpers#haml_buffer` to achieve any meaningful performance difference, but it deserves a separate pull request.


k0kubun:
re-created at https://github.com/haml/haml/pull/1057

PR #1057, opened 2021-06-06
Drop Haml helpers

k0kubun:
I was asked by @hcatlin to bring Hamlit's speed to Haml, breaking backward compatibility as needed. This PR demonstrates what it _at least_ takes. You cannot use `= preserve` or `= find_and_preserve`, but still `~` works and it internally calls `Haml::Helpers.find_and_preserve`.

We also need to get rid of `Haml::Helpers#haml_buffer` to achieve any meaningful performance difference, but it deserves a separate pull request.


hcatlin:
@k0kubun Thanks for this- maybe what we should consider is to mark these as deprecated in the Haml 5 main branch and throw warnings... then that might make an easy transfer for people.

And, I wasn't thinking of changing the Haml code--- but just wholesale replacing it with Hamlit. :P 

Would need to work with the `haml-rails` team if we took that approach, but it might be easier and quicker.

k0kubun:
> And, I wasn't thinking of changing the Haml code--- but just wholesale replacing it with Hamlit. :P
>
> Would need to work with the haml-rails team if we took that approach, but it might be easier and quicker.

It's true that rewriting the entire project could be easy. TBH I felt https://github.com/haml/haml/pull/1058 is more difficult than I expected.

My concern, however, is that I suspect some other gems we don't control could be also broken. If we replace Haml with Hamlit, it automatically means we break every single API that haml.gem has. So Haml integration in all existing Tilt versions would stop working (though Hamlit itself has Tilt integration unlike Haml, so we might be able to do something to force it to be used though). I hope things like haml-lint won't be impacted because Hamlit shares the same parser, but there's no guarantee.

So we need to choose either API backward compatibility + large development effort (use this PR and https://github.com/haml/haml/pull/1058) or stability in Rails integration + small development effort (use Hamlit).

hcatlin:
I didn't really think about Tilt integration.....

k0kubun:
> And, I wasn't thinking of changing the Haml code--- but just wholesale replacing it with Hamlit. :P

Prepared a patch for it https://github.com/haml/haml/pull/1059.

> I didn't really think about Tilt integration.....

Now I took a look at the current integration in Tilt, and it turned out that we actually branched it for Haml 5 support :stuck_out_tongue: https://github.com/rtomayko/tilt/blob/v2.0.10/lib/tilt/haml.rb#L10-L11. We could do that again as a last resort, but looking at the code, I think we could also define `Haml::TempleEngine` alias from the new `Haml::Engine` (former `Hamlit::Engine`) and define `#precompiled_with_ambles` again to support it. So Tilt integration is probably not a problem.

`Haml::Parser` users like haml-lint.gem also have a similar problem because `Hamlit::Parser` wasn't compatible with `Haml::Parser`, but because Hamlit embedded a duplicate of `Haml::Parser`, we should actually refactor that part once it's merged. Then it will not be an issue.

I'll wait for your decision about which direction to take.

k0kubun:
Seems like this has no attention anymore, closing.

PR #1060, opened 2021-06-25
Authorize `@` and `#` characters in attribute name

seballot:
Those characters are valid attributes name, and are used for example in VueJs


hcatlin:
@seballot can you add tests to cover these?

hcatlin:
@seballot Hey Man! Sorry for having to do this, but we are changing the core of the Haml engine over to a hamlit-based one. We're not doing more development on the 5.x branch.

But THANK YOU for trying to contribute <3

PR #1090, opened 2022-09-26
Remove mention of `Haml::Engine` in favor of `Haml::Template`

danieldiekmeier:
In Haml 6, the suggested code for using Haml programmatically was changed from `Haml::Engine` to `Haml::Template`. This PR adjusts the mention of `Haml::Engine` in the line before to be in line with the example code.


k0kubun (approved):
Thanks!

PR #1093, opened 2022-09-28
Fallback to Ruby instead of C ext in util.rb also

kateinoigakukun:
This is a follow-up commit for 8f29bed6ae292bcef8b82cc2baf99a423e2d6191


k0kubun (approved):
Thanks!

PR #1095, opened 2022-09-29
Explain special case of `data-` and `aria-` attributes

danieldiekmeier:
In the updated documentation, it is explained that only the `Haml::AttributeBuilder::BOOLEAN_ATTRIBUTES` are treated as boolean attributes. While testing this, I noticed that this also still works for `data-` and `aria-` attributes.

This makes sense, since this is exactly what the code does:

https://github.com/haml/haml/blob/b3cdafdaf09433af4b0988a533414d9409e72f2a/lib/haml/attribute_compiler.rb#L55

So I think it might be a good addition to the docs to mention this explicitely.

Thanks for your consideration!

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -294,7 +294,19 @@ or using `true` and `false`:
     %input(selected=true)
 
 This feature works only for attributes that are included in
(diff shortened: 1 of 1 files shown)


k0kubun (approved):
:heart_eyes:

PR #1144, opened 2023-09-15
add default_format method to RailsTemplate for Turbo compatibility

lazylester:
as discussed in 'issues'

### DIFF
--- a/lib/haml/rails_template.rb
+++ b/lib/haml/rails_template.rb
@@ -48,6 +48,11 @@ module Haml
       Engine.new(options).call(source)
     end
 
+    # Rails Turbo looks for this
+    def default_format
+      :html
+    end
+
     def supports_streaming?
       RailsTemplate.options[:streaming]
     end


k0kubun:
Thanks!

timdiggins:
I think this change should be rolled back for the following reasons:

* Not all uses of Haml are to create html content-type responses and this can cause issues for them (e.g. js or js with inline html).

* It seems like dhh's comments on the original issue in turbo-rails just before he closed it (https://github.com/hotwired/turbo-rails/issues/287#issuecomment-1133911889) were to say that the naming convention (including .html. in view template names) is required. 
* I don't think Haml should have code that is there to just make Turbo-rails work for people who don't want to follow this convention. And it's pretty easy to rename all your xxx.haml files to xxx.html.haml files - or to add a monkey patch in your app.

NB: "default_format" is quite underdocumented and seems confusingly named - when you define it you prevent the actual default fallback format (which is empty) from being used ( see `ActionView::TemplateDetails::Requested.format_or_default`)

lazylester:
@timdiggins haml has historically always supported `#default_format`. If it's considered appropriate to remove this support, as happened when hamlit was merged, then it should first be deprecated, so as to smooth the path for developers. 

If restoring support for '#content_type` is causing problems in some scenarios (I'm not at all clear on the js scenario you mention), then it must always have been problematic, surely?

Whilst DHH may assert that inclusion of .html in view templates is required, it actually never has been a requirement, but simply common practice.

k0kubun:
Can either of you work on a pull request? As long as it passes tests, I'm open to merging one.

lazylester:
good research @timdiggins.

I think there's a third option, going forwards, which is to deprecate the use of `#default_format` and announce the intention to remove it in a later version. Really this has been my point all along, that behaviour shouldn't just be dropped silently.

timdiggins:
@lazylester unless I misunderstood, your third option seems to refer to how rails should work, not how haml works? Using the template default_format is a novelty in Haml 5 / 6, so "deprecating it" doesn't make sense I think (why introduce something in a patch release and then deprecate it).

I think the monkey patch approach would still work for people using Turbo-Rails (just only with RailsTemplate and without the Plugin pathc) and wouldn't cause any issues with existing Haml users who aren't using turbo (or not yet, or not exclusively).

Hence why I'm suggesting reverting this. Thoughts Les, or have I got the wrong end of the stick?

lazylester:
The monkey patch works just fine. The problem I'm addressing, and the one I experienced, is that my app broke when I upgraded to the hamlit version of haml, because a feature of haml had been dropped. So I had to spend some time troubleshooting the internals of rails turbo to figure out why. Since it appears to me to be a pretty benign feature to restore, it seemed like an easy way to reduce the migration pain for other developers. 

All your forensic work is interesting, but it doesn't address the problem that I, and evidently others too (from the rails issue discussion) have encountered.

timdiggins:
> All your forensic work is interesting, but it doesn't address the problem that I, and evidently others too (from the rails issue discussion) have encountered.

Ah, ok, yes, I wasn't very clear! My suggestion is to revert this commit, but also as part of that PR noting the requirement for turbo-rails compatibility in either the documentation or in the CHANGLEOG for Haml 6 breaking changes something like the following:

> For people using Turbo-rails and Haml 6+ need to either:
> *  follow the naming convention with the html format (ie. ensure any `.haml` files end `.html.haml`), or
> *  add a monkey patch as follows:
> 
>         # config/initializers/haml.rb
>         require "haml/rails_template"
>       
>         module Haml
>           class RailsTemplate
>             def default_format
>               :html
>             end
>           end
>         end

@lazylester What do you think? Would you agree to this change? If so, I'm happy to contribute the PR if you don't want to. But I wouldn't want to contribute the reversion PR without your agreement.

lazylester:
documenting the problem is good, but not as good as a deprecation notice. But I'm still not understanding why inclusion of the `default_format` method is problematic. A js file with inline html, you said. Wasn't this problematic for haml prior to its merging with hamlit, when `default_format` was supported? Is there some framework other than turbo that is testing the value of `default_format` and getting the wrong result?

k0kubun:
Thank you both, @timdiggins and @lazylester, for your thoughtful discussion on this issue.

> haml has historically always supported #default_format
>
> I wasn't aware of the previous use of default_format in haml (pre haml 5 I'm guessing, but I haven't found a def default_format in v3 or v4)
>
> Using the template default_format is a novelty in Haml 5 / 6
>
> Wasn't this problematic for haml prior to its merging with hamlit, when default_format was supported?

Let me get this straight. Here's the `template_handler` registered to Rails in v4 (no influence of Hamlit, which should have accumulated all historical code) and v5 (not entirely replaced by Hamlit yet, but migrated to Temple):

* v4.0.7: [Haml::Plugin](https://github.com/haml/haml/blob/4.0.7/lib/haml/template/plugin.rb)
* v5.2.2: [Haml::Plugin](https://github.com/haml/haml/blob/v5.2.2/lib/haml/plugin.rb)

Neither of these defines `#default_format`, at least inside Haml. So, to my knowledge, haml.gem defining `#default_format` is completely a new feature in v6.1.3.

---

Since this is a fairly new feature, I believe reverting it is also fixes https://github.com/haml/haml/issues/1152 and https://github.com/haml/haml/issues/1154, and this was not consistent with the recommendation by DHH himself, I'll revert this as @timdiggins proposed, in a minor release as a semi-breaking bugfix.

To mitigate the impact for Turbo users, I'll cut a teeny release before that which only prints a deprecation warning in `#default_format` as @lazylester proposed, and commit https://github.com/haml/haml/pull/1144#issuecomment-1759450466 as a documentation.

PR #1146, opened 2023-09-27
Drop C extension

k0kubun:
* One of the main advantages of the Haml C extension was its HTML escape implementation, but I ported its optimization to `ERB::Escape` in Ruby 3.2 and now Haml uses it https://github.com/haml/haml/pull/1145.
* The rest of the C extension exists for rendering attributes, but in most cases, the Haml compiler simplifies generated code and removes the need for the C extension. So it's not called a lot or important in the first place.
* YJIT is getting better in recent versions. If you run this Ruby version with YJIT, the speedup obtained by writing this in C becomes smaller than running it without YJIT. I'd like to invest in YJIT's performance by turning Haml's attribute builder into a potential optimization target for YJIT.
* It has some `rb_funcall` calls. As we optimize YJIT further, YJIT-optimized Ruby code might potentially outperform this C extension since Ruby -> C -> Ruby calls are somewhat expensive by nature. 
* It's tedious to run `rake release` twice for every release.


agrobbin:
@k0kubun I think this might have introduced an issue with the handling of a hash value for `:aria`. On haml 6.1, this:

```haml
%div{ aria: { label: 'foo' } }
```

Would compile to this:

```html
<div aria-label='foo'></div>
```

Now, on 6.2, it compiles to this:

```html
<div aria='{:label=&gt;"foo"}'></div>
```

I *think* it's due to `aria` being missing [here](https://github.com/haml/haml/pull/1146/files#diff-e16177d76c0c7fb4afbaee125b89c466cc31eeca58e86f2f1f3708d81e1e298cR25-R26), but accounted for in the C extension.

k0kubun:
Can you prepare a repository that reproduces it? I cannot reproduce your issue locally. We also have a test case similar to it https://github.com/haml/haml/blob/v6.2.0/test/haml/engine/old_attribute_test.rb#L387-L393 and it's passing. 

```rb
$ irb -rhaml
irb(main)[01:0]> RUBY_DESCRIPTION
=> "ruby 3.2.2 (2023-03-30 revision e51014f9c0) [x86_64-linux]"
irb(main)[02:0]> Haml::VERSION
=> "6.2.0"
irb(main)[03:0]> Haml::Template.new { "%div{ aria: { label: 'foo' } }" }.render
=> "<div aria-label='foo'></div>\n"
```

k0kubun:
OK, I managed to make a reproducible issue from the code you pointed out:

```rb
$ irb -rhaml
irb(main)[01:1]" Haml::Template.new { <<~EOS }.render
irb(main)[02:1]"   - hash = { aria: { label: 'foo' } }
irb(main)[03:1]"   %div{ hash }
irb(main)[04:0]> EOS
=> "<div aria='{:label=&gt;&quot;foo&quot;}'></div>\n"
```

I don't think the `%div{ aria: { label: 'foo' } }` case is broken, but I can fix that one instead.

k0kubun:
Fixed in v6.2.1.

k0kubun:
Fixed another `aria`-related issue in v6.2.2.

agrobbin:
Whoops, yes that is actually exactly what my templates are doing (technically `%div{ **hash }`), when I reduced it down to it's simplest version, I went a step too far. Thanks for the quick work here regardless!

PR #1155, opened 2023-11-23
Add support for case-in statements

jdelStrother:
WDYT to adding case-in support?

```haml
- case type_and_details
- in [:user_account, user]
  Profile #{user.name}
- in [:post, post]
  Post #{post.title}

```
Let me know if you think it needs more tests. I looked at adding some to engine_tests.rb, but they seem more like regression tests, wasn't sure of the best place to add it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,12 +146,6 @@ Once this is complete, you should be able to run the test suite:
 rake
 ~~~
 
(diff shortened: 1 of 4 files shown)


k0kubun (approved):
Nice, thanks!

PR #1186, opened 2025-06-04
Remove newlines from template annotation comments

CyberAP:
This change allows for proper display of the HTML when `annotate_rendered_view_with_filenames` option is used.

Right now this option provides incorrect HTML output in development: adding new lines after each of the HTML comments.

This results in a UI different from production where these comments are not included.

Because `\n` is preserved as whitespace in HTML it can cause issues in inline layout.

### DIFF
--- a/lib/haml/rails_template.rb
+++ b/lib/haml/rails_template.rb
@@ -40,8 +40,8 @@ module Haml
 
       if ActionView::Base.try(:annotate_rendered_view_with_filenames) && template.format == :html
         options = options.merge(
-          preamble: "<!-- BEGIN #{template.short_identifier} -->\n",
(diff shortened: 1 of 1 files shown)


CyberAP:
@amatsuda hi! Maybe you would have some time to review this tiny fix if at all possible?

It's causing visual artefacts when using inline layout in the development mode.

| Current | With patch applied |
|-|-|
| image | image |

Here's the [Rails configuration of this for ERubi](https://github.com/rails/rails/blob/4b7fb1d14b954d2348d2065ff955466479509656/actionview/lib/action_view/template/handlers/erb.rb#L89-L90) for comparison. You can see that the new line is not there.

PR #1205, opened 2026-03-31
Improve frozen strings management

tagliala:
Hello, I was undecided if submit this PR or not, because I couldn't find a benchmark proving actual gain

I'm also unsure if the binary semantics statement in the commit message is correct (there is a "should")

---

This commit replaces occurrences of `''.dup` and `String.new` with `+''`, and avoids using `+""` for interpolated strings.

This change is motivated by both performance and clarity: on Ruby 3.2 (our minimum version), `+''` is significantly faster than `''.dup` or `String.new`.

On 3.3 and later, `+''` and `''.dup` are equivalent, so this is a forward-compatible improvement.

Using `+""` on interpolated strings is also unnecessary, as interpolated strings are mutable by default on Ruby 3.0 and later.

Note on safety: `String.new` yields an ASCII-8BIT (BINARY) empty string, while `+''` yields UTF-8. Haml normalizes template encodings and defaults outputs to UTF-8 (see Util.check_haml_encoding and tests like EngineTest#test_default_encoding). The `String.new` instances touched are not used for byte-level binary processing and do not rely on BINARY semantics. Given Haml's UTF-8 orientation, this change should be safe.

These changes were made automatically with:

```
rubocop --plugin rubocop-performance --only Performance/UnfreezeString lib/**/*.rb -A
rubocop --only Style/RedundantInterpolationUnfreeze lib/**/*.rb -a
```

### DIFF
--- a/lib/haml/parser.rb
+++ b/lib/haml/parser.rb
@@ -227,7 +227,7 @@ module Haml
       end
 
       def inspect
-        %Q[(#{type} #{value.inspect}#{children.each_with_object(''.dup) {|c, s| s << "\n#{c.inspect.gsub!(/^/, '  ')}"}})].dup
+        %Q[(#{type} #{value.inspect}#{children.each_with_object((+'')) {|c, s| s << "\n#{c.inspect.gsub!(/^/, '  ')}"}})]
       end
     end
 
@@ -397,7 +397,7 @@ module Haml
     def haml_comment(text)
       if filter_opened?
         @flat = true
-        @filter_buffer = String.new
+        @filter_buffer = +''
         @filter_buffer << "#{text}\n" unless text.empty?
         text = @filter_buffer
         # If we don't know the indentation by now, it'll be set in Line#tabs
@@ -532,7 +532,7 @@ module Haml
 
       if filter_opened?
         @flat = true
-        @filter_buffer = String.new
+        @filter_buffer = +''
         # If we don't know the indentation by now, it'll be set in Line#tabs
         @flat_spaces = @indentation * (@template_tabs+1) if @indentation
       end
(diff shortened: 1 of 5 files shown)


lib/haml/parser.rb
```diff
@@ -774,7 +774,7 @@ def parse_new_attribute(scanner)
 
       return name, [:static, content.first[1]] if content.size == 1
       return name, [:dynamic,
-        %!"#{content.each_with_object(''.dup) {|(t, v), s| s << (t == :str ? Util.inspect_obj(v)[1...-1] : "\#{#{v}}")}}"!]
+        %!"#{content.each_with_object((+'')) {|(t, v), s| s << (t == :str ? Util.inspect_obj(v)[1...-1] : "\#{#{v}}")}}"!]
```
k0kubun: Do we really need parens for this?

```suggestion
        %!"#{content.each_with_object(+'') {|(t, v), s| s << (t == :str ? Util.inspect_obj(v)[1...-1] : "\#{#{v}}")}}"!]
```

lib/haml/parser.rb
```diff
@@ -227,7 +227,7 @@ def initialize(*args)
       end
 
       def inspect
-        %Q[(#{type} #{value.inspect}#{children.each_with_object(''.dup) {|c, s| s << "\n#{c.inspect.gsub!(/^/, '  ')}"}})].dup
+        %Q[(#{type} #{value.inspect}#{children.each_with_object((+'')) {|c, s| s << "\n#{c.inspect.gsub!(/^/, '  ')}"}})]
```
k0kubun:
```suggestion
        %Q[(#{type} #{value.inspect}#{children.each_with_object(+'') {|c, s| s << "\n#{c.inspect.gsub!(/^/, '  ')}"}})]
```

ditto

tagliala:
Rebased and force-pushed this branch. After the latest parser-related commit, the change is now smaller.

k0kubun (approved):

PR #1210, opened 2026-06-07
Remove obsolete TruffleRuby compatibility skips

ollym:
Removes the historical TruffleRuby short circuits now that TruffleRuby 34 supports the Ripper APIs Haml relies on.

Changes:

- Removed TruffleRuby-specific Ripper.lex / Ripper.sexp runtime guards.
- Removed TruffleRuby-only test skips across parser, compiler, Rails template, interpolation, and optimization tests.
- Enabled stackprof on TruffleRuby by removing the Gemfile exclusion.

Verification:

LANG=en_US.UTF-8 RBENV_VERSION=truffleruby-34.0.1 bundle exec rake test

Result:

1093 runs, 1110 assertions, 0 failures, 0 errors, 199 skips

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,7 +20,7 @@ if /java/.match?(RUBY_PLATFORM) # JRuby
 else
   gem 'redcarpet'
 
-  if RUBY_PLATFORM !~ /mswin|mingw/ && RUBY_ENGINE != 'truffleruby'
+  if RUBY_PLATFORM !~ /mswin|mingw/
     gem 'stackprof'
   end
 end
--- a/lib/haml/attribute_parser.rb
+++ b/lib/haml/attribute_parser.rb
@@ -8,8 +8,7 @@ module Haml
 
     # @return [TrueClass, FalseClass] - return true if AttributeParser.parse can be used.
     def self.available?
-      # TruffleRuby doesn't have Ripper.lex
-      defined?(Ripper) && Ripper.respond_to?(:lex) && Temple::StaticAnalyzer.available?
+      Temple::StaticAnalyzer.available?
     end
 
     def self.parse(text)
--- a/lib/haml/compiler/script_compiler.rb
+++ b/lib/haml/compiler/script_compiler.rb
@@ -21,10 +21,6 @@ module Haml
       end
 
       def compile(node, &block)
-        unless Ripper.respond_to?(:lex) # No Ripper.lex in truffleruby
-          return dynamic_compile(node, &block)
-        end
-
         no_children = node.children.empty?
         case
         when no_children && node.value[:escape_interpolation]
--- a/lib/haml/compiler/tag_compiler.rb
+++ b/lib/haml/compiler/tag_compiler.rb
@@ -28,10 +28,8 @@ module Haml
           nil
         when node.value[:parse]
           return compile_interpolated_plain(node) if node.value[:escape_interpolation]
-          if Ripper.respond_to?(:lex) # No Ripper.lex in truffleruby
-            return delegate_optimization(node) if RubyExpression.string_literal?(node.value[:value])
-            return delegate_optimization(node) if Temple::StaticAnalyzer.static?(node.value[:value])
-          end
+          return delegate_optimization(node) if RubyExpression.string_literal?(node.value[:value])
+          return delegate_optimization(node) if Temple::StaticAnalyzer.static?(node.value[:value])
 
           var = @identity.generate
           [:multi,
--- a/lib/haml/string_splitter.rb
+++ b/lib/haml/string_splitter.rb
@@ -1,139 +1,129 @@
 # frozen_string_literal: true
-begin
-  require 'ripper'
-rescue LoadError
-end
+require 'ripper'
 
 module Haml
   # Compile [:dynamic, "foo#{bar}"] to [:multi, [:static, 'foo'], [:dynamic, 'bar']]
   class StringSplitter < Temple::Filter
-    if defined?(Ripper) && Ripper.respond_to?(:lex)
-      class << self
-        # `code` param must be valid string literal
-        def compile(code)
-          [].tap do |exps|
-            tokens = Ripper.lex(code.strip)
-            tokens.pop while tokens.last && [:on_comment, :on_sp].include?(tokens.last[1])
+    class << self
+      # `code` param must be valid string literal
+      def compile(code)
+        [].tap do |exps|
+          tokens = Ripper.lex(code.strip)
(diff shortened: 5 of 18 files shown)


eregon:
Nice cleanups :) And my apologies you had to workaround in `haml` for the lack of/incomplete `Ripper` in TruffleRuby.

andrykonchin:
Is there a chance this PR could be reviewed?

I am interested in the changes to `parse_old_attributes`. The current code relies on a Ripper quirk (emitting `:on_embexpr_end` for unbalanced braces) to detect the end of hash-like attributes. This quirk is not supported by Prism's Ripper emulation, which TruffleRuby relies on.

Switching to a manual brace count as done in this PR fixes this compatibility issue.

cc @eregon

ollym:
@k0kubun ?

eregon:
For context [we tried](https://github.com/ruby/prism/pull/4164) to change `Prism::Translation::Ripper` to behave closer to ::Ripper for this case but that was [considered too hacky](https://github.com/ruby/prism/pull/4164#issuecomment-4946764574) because it relies on what RIpper does on invalid syntax which is rather unpredictable and inconsistent.

lib/haml/parser.rb
```diff
@@ -686,10 +683,8 @@ def parse_old_attributes(text)
         # Old attributes often look like a valid Hash literal, but it sometimes allow code like
         # `{ hash, foo: bar }`, which is compiled to `_hamlout.attributes({}, nil, hash, foo: bar)`.
         #
-        # To scan such code correctly, this scans `a( hash, foo: bar }` instead, stops when there is
-        # 1 more :on_embexpr_end (the last '}') than :on_embexpr_beg, and resurrects '{' afterwards.
-        balanced, rest = balance_tokens(text.sub(?{, METHOD_CALL_PREFIX), :on_embexpr_beg, :on_embexpr_end, count: 1)
-        attributes_hash = balanced.sub(METHOD_CALL_PREFIX, ?{)
+        # Balance Ripper's brace tokens so `{ a: "}" }` is handled correctly.
+        attributes_hash, rest = balance_tokens(text, :on_lbrace, :on_rbrace)
```
andrykonchin: Should we handle also `:on_tlambeg` (for a lambda expression)?

```
Ripper.lex('-> {}')
=>
[[[1, 0], :on_tlambda, "->", ENDFN],
 [[1, 2], :on_sp, " ", ENDFN],
 [[1, 3], :on_tlambeg, "{", BEG],
 [[1, 4], :on_rbrace, "}", END]]
```

k0kubun: Why did this need to be changed? It doesn't seem relevant to whether `Ripper.lex` is available or not.

k0kubun: oh that's what https://github.com/haml/haml/pull/1210#issuecomment-4947484906 is about.

k0kubun:
This PR does more than "Remove obsolete TruffleRuby compatibility skips" in `lib/haml/parser.rb`, so I'm not comfortable merging it right away (see also: https://github.com/haml/haml/pull/1210#discussion_r3590420271). Can you separate a PR for that change if possible?

> The current code relies on a Ripper quirk (emitting :on_embexpr_end for unbalanced braces) to detect the end of hash-like attributes. This quirk is not supported by Prism's Ripper emulation, which TruffleRuby relies on.

I'm not sure if I follow. My understanding is that CRuby would also use Prism to emulate Ripper unless `--parser=parse.y` is specified, but are you saying `Ripper.lex` is still implemented by `parse.y` even with `--parser=prism`?

lib/haml/parser.rb
```diff
@@ -686,10 +683,8 @@ def parse_old_attributes(text)
         # Old attributes often look like a valid Hash literal, but it sometimes allow code like
         # `{ hash, foo: bar }`, which is compiled to `_hamlout.attributes({}, nil, hash, foo: bar)`.
         #
-        # To scan such code correctly, this scans `a( hash, foo: bar }` instead, stops when there is
-        # 1 more :on_embexpr_end (the last '}') than :on_embexpr_beg, and resurrects '{' afterwards.
-        balanced, rest = balance_tokens(text.sub(?{, METHOD_CALL_PREFIX), :on_embexpr_beg, :on_embexpr_end, count: 1)
```
k0kubun: Why did you drop `METHOD_CALL_PREFIX`? Haml's old attributes (`%a{}`) is not a Hash literal. It works like a method argument (`%a{ hash, foo: bar }`), so removing it makes it an invalid Ruby syntax to be parsed. Even though we're feeding the code to a lexer, not a parser, I'm not sure if it's reliable.

lib/haml/parser.rb
k0kubun: With regard to this PR, https://github.com/haml/haml/pull/1210#discussion_r3590420271 should be resolved first.

eregon:
> I'm not sure if I follow. My understanding is that CRuby would also use Prism to emulate Ripper unless `--parser=parse.y` is specified, but are you saying `Ripper.lex` is still implemented by `parse.y` even with `--parser=prism`?

Yes it is, it can be surprising but that's the case.

k0kubun:
OK. I'm fine with dropping the Ripper-specific token until we find an issue with it.

I'll wait for https://github.com/haml/haml/pull/1210#discussion_r3590420271 to be addressed though. It has nothing to do with whether `Ripper.lex` is powered by Prism or parse.y.

lib/haml/parser.rb
andrykonchin: Yeah, I see. We should preserve the trick with `METHOD_CALL_PREFIX`.

Extracted the fix into a separate PR #1212 and described the problem and the solution. Could you please take a look?

eregon: > so removing it makes it an invalid Ruby syntax to be parsed.

Wasn't it already invalid before?
I think it was, given the comment mentioning it parsed `a( hash, foo: bar }`.

Maybe we should convert the last `}` to a `)` then it would be a real method call?
I guess the trouble is it's not trivial to find that one, hence the balancing to find it.

k0kubun: > Wasn't it already invalid before?
> I think it was, given the comment mentioning it parsed a( hash, foo: bar }

`count: 1` is there to deal with that problem. Since we're lexing it instead of parsing it, it shouldn't be a problem until you lex the last `}` token. It probably parses a valid syntax just fine, but converting the last `}` to `)` might be helpful for better error messages.

Merged https://github.com/haml/haml/pull/1212. Would you mind rebasing this PR?

ollym:
@k0kubun done, sorry for the delay

k0kubun (approved):

PR #1211, opened 2026-07-14
Skip Ripper with cheap regex prefilters for static?/string_literal?

wilfison:
# Skip Ripper with cheap regex prefilters for `static?` / `string_literal?`

## Summary

During compilation, Haml asks two predicates whether a Ruby snippet is a **static expression** (`Temple::StaticAnalyzer.static?`) or a **string literal** (`Haml::RubyExpression.string_literal?`). Both run a **full Ripper parse/lex** for _every_ snippet — every `= expr`, every inline tag value, and every `[:dynamic]` node in the tree — even for expressions that obviously can't be static or literal, like `render partial: 'x'` or `@user.name`.

This PR puts a **cheap regex gate on the first token** in front of each predicate. Ripper still runs, but only for snippets that could plausibly pass. The generated Ruby is **byte-for-byte identical** to before; the change is purely about avoiding wasted parses.

Measured **~15–22% faster template compilation** (details below), with no change to rendered output.

## How it works

A static expression can only begin with a literal (`"`/`'`/`[`/`(`/`{`/`:`/digit/`%w`…) or `true`/`false`/`nil`. A string literal can only begin with a quote or `%q`/`%Q`. That's a **necessary condition**, so:

- **Regex doesn't match** → the snippet provably can't be static/literal → return `false` immediately, no Ripper.
- **Regex matches** → fall through to the existing Ripper path (unchanged).

A false positive (regex matches but Ripper disagrees) is harmless — it just takes the old path. There are no false negatives for the tokens these predicates accept, so nothing that used to fold to `[:static, …]` stops folding. Even hypothetically, a missed fold would only skip a compile-time optimization and emit `[:dynamic, …]`, which renders to the identical string — never a correctness change.

```ruby
# Haml::RubyExpression
STRING_LITERAL_START = /\A\s*(?:["']|%[qQ]?[({\[|!<~])/

# Haml::StaticAnalyzerPrefilter (prepended onto Temple::StaticAnalyzer)
STATIC_START = /\A\s*(?:["'\[({:\d]|%[qQwWiI]?[({\[|!<~]|true[\s)\]}]*\z|false[\s)\]}]*\z|nil[\s)\]}]*\z)/
```

### Secondary cleanups (same area)

- **Dropped a redundant parse.** `RubyExpression.string_literal?` used to call `syntax_error?` (parse 1) and then `Ripper.sexp` (parse 2). But `Ripper.sexp` already returns `nil` on a syntax error, so the first parse was redundant — removed.
- **Removed duplication.** `StringSplitter` carried its own private copy of `string_literal?` plus a duplicate `SyntaxChecker < Ripper` class. It now delegates to `Haml::RubyExpression.string_literal?`, inheriting the prefilter for free (~25 fewer lines).

## Safety / correctness

- **Byte-identical output.** Compiling the entire `benchmark/**/*.haml` corpus produces the same checksum as `main` `c910bca690dd0749`.
- **Full suite green:** `bundle exec rake test` → `1096 runs, 0 failures, 0 errors` (existing `1093` + 3 new prefilter cases).
- New unit tests cover the fast-reject path (`render partial: 'x'`, `@user.name`, `foo('bar')`).

## Benchmarks

Environment: Ruby 4.0.2, single machine. Compile-time only (this is a compile-time change; rendering is unaffected). Run-to-run variance is non-trivial on this box, so the aggregate figure is reported as median-of-5 plus best-of-run.

### Whole corpus (aggregate compile time, lower is better)

21 templates, 200 compile passes per run, 5 runs:

| branch            | ms/pass (median) | ms/pass (best) |
| ----------------- | ---------------: | -------------: |
| `main` (baseline) |            41.98 |          39.04 |
| this PR           |            32.83 |          32.41 |
| **improvement**   |       **-21.8%** |     **-17.0%** |

### Per template (`rake bench COMPILE=1`, benchmark-ips, higher i/s is better)

| template                                  | `main` (i/s) | this PR (i/s) | throughput |               compile time |
| ----------------------------------------- | -----------: | ------------: | ---------: | -------------------------: |
| `benchmark/etc/real_sample.haml`          |         41.6 |          49.0 | **+17.6%** | 24.01ms → 20.42ms (-15.0%) |
| `benchmark/etc/string_interpolation.haml` |         2028 |          2238 |     +10.4% |  0.493ms → 0.447ms (-9.3%) |
| `benchmark/script.haml`                   |         2027 |          2228 |      +9.9% |  0.493ms → 0.449ms (-8.9%) |
| `benchmark/etc/static_analyzer.haml`      |         1971 |          2107 |      +6.9% |  0.507ms → 0.475ms (-6.3%) |

The realistic full template (`real_sample`) benefits most (~+17%), matching the aggregate; the tiny single-purpose templates gain less because a larger share of their time is spent in other pipeline stages.

## How the benchmarks were run

Baseline and PR are compared by checking out each branch and running the identical measurement (the working tree is clean, so `git checkout` swaps only the code):

```sh
# Aggregate corpus compile time (median-of-5), self-contained:
run() {
  ruby -Ilib -e '
    require "haml"; require "digest"
    t = Dir["./benchmark/**/*.haml"].sort.map { |f| File.read(f) }
    d = Digest::SHA256.new; t.each { |s| d << Haml::Engine.new.call(s) }
    2.times { t.each { |s| Haml::Engine.new.call(s) } }            # warmup
    s = 5.times.map { GC.start
      a = Process.clock_gettime(Process::CLOCK_MONOTONIC)
      200.times { t.each { |x| Haml::Engine.new.call(x) } }
      (Process.clock_gettime(Process::CLOCK_MONOTONIC) - a) * 1000 / 200 }.sort
    puts "checksum #{d.hexdigest[0,16]}  median %.3f ms/pass  best %.3f" % [s[2], s[0]]'
}
git checkout main                 && run
git checkout perf/ripper-prefilters && run

# Per-template (official tool):
COMPILE=1 TEMPLATE=benchmark/etc/real_sample.haml bundle exec rake bench   # read the "Compilation:" i/s
```

## Notes

- **Temple monkeypatch.** `static?` lives in `Temple::StaticAnalyzer`, and it's also used internally by the engine's `filter :StaticAnalyzer`, so the prefilter is applied via `Temple::StaticAnalyzer.singleton_class.prepend(...)`. A cleaner long-term fix would be to upstream the prefilter into Temple.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Haml Changelog
 
+## Unreleased
+
+* Skip Ripper parsing with cheap regex prefilters when compiling static expressions and string literals https://github.com/haml/haml/pull/1211
+
 ## 7.2.2
 
 * Remove obsolete TruffleRuby compatibility skips https://github.com/haml/haml/pull/1210
--- a/lib/haml/ruby_expression.rb
+++ b/lib/haml/ruby_expression.rb
@@ -1,9 +1,13 @@
 # frozen_string_literal: true
 require 'ripper'
+require 'temple/static_analyzer'
 
 module Haml
   class RubyExpression < Ripper
     class ParseError < StandardError; end
+
+    # Necessary condition for a string literal: a quote or %q/%Q. Cheap gate before Ripper.
+    STRING_LITERAL_START = /\A\s*(?:["']|%[qQ]?[({\[|!<~])/
 
     def self.syntax_error?(code)
       self.new(code).parse
@@ -13,8 +17,9 @@ module Haml
     end
 
     def self.string_literal?(code)
-      return false if syntax_error?(code)
+      return false unless STRING_LITERAL_START.match?(code)
 
+      # Ripper.sexp already returns nil on syntax errors, so no separate syntax_error? check.
       type, instructions = Ripper.sexp(code)
       return false if type != :program
       return false if instructions.size > 1
@@ -29,4 +34,16 @@ module Haml
       raise ParseError
     end
   end
+
+  # Necessary condition for a static expression. A non-match skips Temple's Ripper parse.
+  module StaticAnalyzerPrefilter
+    STATIC_START = /\A\s*(?:["'\[({:\d]|%[qQwWiI]?[({\[|!<~]|true[\s)\]}]*\z|false[\s)\]}]*\z|nil[\s)\]}]*\z)/
+    def static?(code)
+      return false if code.nil?
+      return false unless STATIC_START.match?(code)
+      super
+    end
+  end
 end
+
+Temple::StaticAnalyzer.singleton_class.prepend(Haml::StaticAnalyzerPrefilter)
--- a/lib/haml/string_splitter.rb
+++ b/lib/haml/string_splitter.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 require 'ripper'
+require 'haml/ruby_expression'
 
 module Haml
   # Compile [:dynamic, "foo#{bar}"] to [:multi, [:static, 'foo'], [:dynamic, 'bar']]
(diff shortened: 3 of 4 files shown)


k0kubun:
I've been meaning to replace Ripper with Prism (about every use of Ripper in general, but more so for this static/string literal check), so I would accept a change like that. Would you mind giving it a shot?

I'm not going to replace it with a Regexp whether it's faster than either of these though. I might want to tweak what we accept in that filter later, but it'd be harder with Regexp when it comes to parsing Ruby code. The rendering performance (the performance of compiled code, which doesn't include compilation performance) is always the top priority, but we don't necessarily give up maintainability for compilation performance.

wilfison:
That makes sense, thanks for explaining the reasoning. Keeping a Regexp in sync with whatever the filter accepts is a real maintenance cost, and I agree it's not worth paying for compilation performance alone. I'm closing this PR.

As for Prism, I don't think I know it well enough to do that replacement properly right now, and I'd rather not take it on than submit something half-baked. I'll spend some time studying it, and if I get comfortable enough I'll come back with a proper attempt.

Issue #1161, opened 2024-05-13
Template broken with TruffleRuby 24.0.1


n-rodriguez:
Hi there! I'm trying to run a Rails application with TruffleRuby 24.0.1 and the rendering is partially broken.

Here a screenshot :
![Screenshot 2024-05-13 at 02-08-04 #{( Haml Util escape_html_safe((_haml_compiler1))) to_s}](https://github.com/haml/haml/assets/3433835/aad7fbe3-e2ae-48c0-90e4-afee82d619db)

The app runs perfectly with Ruby 3.3.1.

I'm running `haml 6.3.0`: 

```sh
nicolas@MacBook-Pro-de-Nicolas:~/PROJECTS/CONCERTO/concerto$ grep haml Gemfile.lock
    haml (6.3.0)
```

Thank you!

HamptonMakes:
Seems like our tests for TruffleRuby have been failing:
https://github.com/haml/haml/actions/runs/9210369754/job/25337217627#step:5:52

dentarg:
Can you test with TruffleRuby-head build? We had similar failures in https://github.com/sinatra/sinatra/issues/2007 and there was a bug https://github.com/ruby/prism/issues/2646 that will be included in a later TruffleRuby release.

n-rodriguez:
@dentarg TruffleRuby 24.1.0 is out and the problem is fixed. Thank you!

PR #1214, opened 2026-08-07
Replace Ripper with Prism

wilfison:
## Summary

This replaces every use of Ripper in `lib/` with [Prism](https://github.com/ruby/prism). After this change `grep -rn Ripper lib/` returns nothing — Ripper still loads transitively through `Temple::StaticAnalyzer`, which is out of scope here (see [Remaining Ripper usage](#remaining-ripper-usage) below).

This follows up on the discussion in #1211, where the suggestion was:

> I've been meaning to replace Ripper with Prism (about every use of Ripper in general, but more so for this static/string literal check), so I would accept a change like that.

The work is split into two commits: the first moves the syntax and string-literal checks, the second moves the three remaining lexing sites.

Beyond the parser swap, the motivation is that Prism is one shared lexer and parser across CRuby, JRuby and TruffleRuby, whereas Ripper's behaviour differs between them. That difference is what produced #1161, the token-shape handling in #1212, and the two JRuby-specific `skip`s this PR deletes.

## What changed

**`Haml::RubyExpression`** no longer subclasses `Ripper`. `syntax_error?` is `Prism.parse_failure?`, and `string_literal?` inspects the AST in a single parse where it previously did a syntax check followed by `Ripper.sexp`. A node counts as a string literal when it is a `StringNode` or an `InterpolatedStringNode` carrying its own opening delimiter. The delimiter check excludes two shapes Prism reports as string nodes but Ripper did not: adjacent concatenation (`"a" "b"`), which has no delimiter wrapping the whole expression, and character literals (`?a`), which have no quotes. Both would otherwise reach `StringSplitter`, which cannot split them. `string_literal_node` was extracted so that rule lives in one place and the literal is parsed once.

All of it parses with Prism's `partial_script` option, because a template is compiled into a method body: `yield`, `break`, `next` and `redo` are valid there even though they are not at the top level of a script, and Prism would otherwise report them as errors. Prism's changelog introduces the option for exactly this, "for parsing things like ERB sources, where you know it will be evaluated in a different context". It does not loosen anything else — an unterminated string is still an error.

**`Haml::StringSplitter.compile`** walks the AST instead of a token stream. That removes `strip_quotes!`, `compile_tokens!`, `shift_balanced_embexpr`, an `eval`, and `escape_quotes`, whose null-character substitution existed only to make re-evaluating a slice of the source safe. The file went from 130 to 66 lines. The source of an interpolation comes from the byte offsets between its `#{` and `}` rather than from the statements it holds, so the whitespace around them survives — `"#{ '}' }"` still yields `[:dynamic, " '}' "]`.

**`Haml::AttributeParser`** walks `Prism::HashNode#elements` instead of counting bracket, brace, paren and interpolation depth to find the commas separating attributes (115 to 76 lines). Keys come from the source between their delimiters rather than from `unescaped`, because Ripper reported raw token text: `{ "a\0b" => 1 }` has to keep spelling the escape the way the template wrote it, or the attribute name changes. Ruby 3.1 value omission (`{ foo: }`) arrives as a `Prism::ImplicitNode` whose `slice` is `"foo:"`, and is mapped to `''` so the existing fallback to `runtime_compile` still fires.

**`Haml::Parser#balance_tokens`** counts Prism tokens and splits the buffer on the offset where it balances. `Prism.lex` emits no whitespace or comment tokens, so the text can no longer be rebuilt by concatenating them; slicing on the offset also fixes the previous `buf.sub(text, '')`, which could match an earlier occurrence of the same text.

## Behaviour changes

Two bugs fall out of the rewrite. No existing test covered either behaviour.

**Interpolating a variable without braces was silently dropped.** `compile_tokens!` only handled `:on_tstring_content` and `:on_embexpr_beg`, so the variable tokens were discarded at compile time:

```ruby
# before
Haml::Engine.new.call(%q[= "#@foo"])
# => "_buf = ''.dup; \n; _buf << (\"\\n\".freeze); _buf"   # the interpolation is gone

# %p= "#@foo"          rendered <p></p>          instead of <p>value</p>
# %p{ title: "#@foo" } rendered title=""         instead of title="value"
```

Plain text was never affected, because `Haml::Util.unescape_interpolation` rewrites `#@foo` into `#{@foo}` before `StringSplitter` sees it — only explicit string literals in scripts and attributes were broken. Worth noting that `Temple::Filters::StringSplitter` has the same bug on released Temple (`compile('"#@foo"')` returns `[]`), and the open Prism port at judofyr/temple#156 fixes it there the same way.

**An escaped delimiter of a percent literal lost its escape.** `= %q{a\}b}` rendered `a\}b` where Ruby's own value is `a}b`. `escape_quotes` swapped the `{`/`}` delimiters for null characters, so `\}` stopped being an escaped delimiter, and `%q` preserves unknown escapes. This one is inseparable from the refactor: using `StringNode#unescaped` fixes it automatically.

## A deliberate non-change

A multi-line old-attribute hash still refuses to compile statically. Today any newline makes the token walker bail, so those hashes always take the runtime path. Prism parses them happily and would silently start optimising them, but doing so drops a `[:newline]` and shifts every `__LINE__` after it:

```
%span{ a: __LINE__,
  b: __LINE__ }= __LINE__
= __LINE__

# today, and with this PR:   <span a="1" b="2">2</span> / 3
# without the guard:         <span a="1" b="1">1</span> / 2
```

The guard is explicit and commented, with a test pinning it, so making the static path newline-correct can be a separate change.

## Deprecation

`Haml::AttributeParser.available?` answered whether Ripper was present. Prism is a hard dependency now, so it is always true and Haml itself no longer asks. It is kept and marked deprecated rather than removed.

## Verification

The full suite passes with 46 new tests and no new skips:

```
1156 runs, 1175 assertions, 0 failures, 0 errors, 199 skips
```

The two skips this PR deletes only ever fired on JRuby, which is why the count is unchanged. New coverage includes the `unbalanced_brackets` contract, which had none — its three expectations live in `engine_test.rb`'s `EXCEPTION_MAP`, but `test_exception_map` is `skip`ped.

Beyond the assertions, I diffed the **compiled output** of every template in `test/haml/templates/` and `benchmark/` across three option sets, against the commit before this branch. All 5616 lines are byte-identical, which is the check that would have caught a silent flip between the static and runtime attribute paths.

Compilation is faster, measured on the same machine with a template exercising both attributes and interpolation:

| | before | after | |
|---|---|---|---|
| `RubyExpression.string_literal?`, non-literal input | 0.154s | 0.051s | 3.0x |
| `Engine#call`, whole template | 2.895s | 2.202s | 24% |

`prism >= 1.1.0` is added to the gemspec — 1.1.0 is where `partial_script` arrived. Every other Prism API used here, including the `Prism.lex` token names, which are the most version-sensitive detail, was verified to exist as far back as 1.0.0.

I also rendered a real Rails application against this branch, which is how the `partial_script` requirement surfaced: a layout doing `%title` over `#{yield(:title)}` compiled fine under Ripper, which only lexed, and failed once the same text was actually parsed. Nothing in this repository's templates uses a jump inside an interpolation, so neither the suite nor the compiled-output diff caught it. There is now a test for it.

Where a caller builds a string literal itself — the two `compile_interpolated_plain` methods and the `:plain` filter — it asks `StringSplitter.try_compile`, which answers nil rather than raising. Ruby that genuinely does not parse, such as a typo inside an interpolation, is then left to the generated code to report against the template, which is what Ripper's purely lexical splitting did, instead of failing the whole compilation.

## Platforms

The whole matrix is green — 3.2, 3.3, 3.4, 4.0, ruby-head, JRuby and TruffleRuby.

JRuby was the part I could not check while writing this, since all three sites now go through the prism gem's non-CRuby backend, and `Prism.lex` in `balance_tokens` is the least-exercised of the APIs used. It passing also means the two deleted `skip`s genuinely pass there rather than merely being absent, which is the evidence that the Ripper differences between implementations are gone. Locally I only had 4.0.1 and 4.0.2 to run against, so 3.2, 3.3 and 3.4 are CI's word.

## Remaining Ripper usage

`Temple::StaticAnalyzer` still requires Ripper, so it keeps loading through `filter :StaticAnalyzer` and the `static?` calls in the attribute and script compilers. There is already an open PR adding a Prism backend to Temple — judofyr/temple#156, by Prism's author — and because it requires prism opportunistically rather than through Temple's gemspec, and Haml now depends on prism directly, merging it would make Temple select Prism automatically with no further change here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,10 @@
 
 ## Unreleased
 
-* Interpolate `#@ivar`, `#$gvar` and `#@@cvar` in string literals instead of dropping them
-* Keep an escaped delimiter of a percent literal, so `= %q{a\}b}` renders `a}b`
-* Deprecate `Haml::AttributeParser.available?`, which is now always true and will be removed in the future
+* Replace Ripper with Prism https://github.com/haml/haml/pull/1214
+  * Interpolate `#@ivar`, `#$gvar` and `#@@cvar` in string literals instead of dropping them
+  * Keep an escaped delimiter of a percent literal, so `= %q{a\}b}` renders `a}b`
+  * Deprecate `Haml::AttributeParser.available?`, which is now always true and will be removed in the future
 
 ## 7.2.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Haml Changelog
 
+## Unreleased
+
+* Interpolate `#@ivar`, `#$gvar` and `#@@cvar` in string literals instead of dropping them
+* Keep an escaped delimiter of a percent literal, so `= %q{a\}b}` renders `a}b`
+* Deprecate `Haml::AttributeParser.available?`, which is now always true and will be removed in the future
+
 ## 7.2.2
 
 * Remove obsolete TruffleRuby compatibility skips https://github.com/haml/haml/pull/1210
--- a/haml.gemspec
+++ b/haml.gemspec
@@ -28,7 +28,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 3.2.0'
 
-  spec.add_dependency 'prism', '>= 1.0.0'
+  spec.add_dependency 'prism', '>= 1.1.0'
   spec.add_dependency 'temple', '>= 0.8.2'
   spec.add_dependency 'thor'
   spec.add_dependency 'tilt'
--- a/haml.gemspec
+++ b/haml.gemspec
@@ -28,6 +28,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 3.2.0'
 
+  spec.add_dependency 'prism', '>= 1.0.0'
   spec.add_dependency 'temple', '>= 0.8.2'
   spec.add_dependency 'thor'
   spec.add_dependency 'tilt'
--- a/lib/haml/attribute_compiler.rb
+++ b/lib/haml/attribute_compiler.rb
@@ -22,9 +22,7 @@ module Haml
 
     def compile(node)
       hashes = []
-      if node.value[:object_ref] != :nil || !AttributeParser.available?
-        return runtime_compile(node)
-      end
+      return runtime_compile(node) if node.value[:object_ref] != :nil
       [node.value[:dynamic_attributes].new, node.value[:dynamic_attributes].old].compact.each do |attribute_str|
         hash = AttributeParser.parse(attribute_str)
         return runtime_compile(node) if hash.nil? || hash.any? { |_key, value| value.empty? }
--- a/lib/haml/attribute_parser.rb
+++ b/lib/haml/attribute_parser.rb
@@ -1,33 +1,40 @@
 # frozen_string_literal: true
-require 'haml/ruby_expression'
+require 'prism'
 
 module Haml
   class AttributeParser
-    class ParseSkip < StandardError
-    end
-
-    # @return [TrueClass, FalseClass] - return true if AttributeParser.parse can be used.
+    # @deprecated Prism is a hard dependency, so this is always true. Haml itself no longer
+    #   asks, and this will be removed in the future.
+    # @return [TrueClass] - return true if AttributeParser.parse can be used.
     def self.available?
-      Temple::StaticAnalyzer.available?
+      true
     end
 
     def self.parse(text)
       self.new.parse(text)
     end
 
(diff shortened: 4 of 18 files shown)


wilfison:
@k0kubun  back in #1211 you said you'd been meaning to replace Ripper with Prism and asked if I'd give it a shot. I answered that I'd rather study it first than hand you something half-baked, and come back with a proper attempt. Here it is 🙂

No rush at all. Take a look whenever it suits you, and I'm glad to reshape or split it if you'd prefer it in smaller pieces.

k0kubun (approved):
Thanks!

Merged and released the temple change as well.

PR #1215, opened 2026-08-09
Cut per-render and per-line overhead in the runtime helpers and parser

wilfison:
Nine self-contained micro-optimizations to the render hot paths (duplicate `to_s` in every dynamic output, attribute classification, attribute key handling, `escape_html_safe`, `find_and_preserve`) and to the parser (regexes rebuilt per line, per interpolation and per attribute value). No public API changes; rendered HTML is byte-identical throughout.

## Changes

One commit each, in this order:

- [x] Stop `EscapeAny` from appending a duplicate `to_s` to every dynamic output
- [x] Classify boolean attributes with a Set instead of a `case`/`when` splat
- [x] Look up Symbol attribute keys with `Symbol#name` instead of `to_s`
- [x] Flatten nested attributes without `Hash#==` or an unconditional `tr`
- [x] Pick the `html_safe` escaper once at load instead of `respond_to?` per value
- [x] Build the preserve regex once per tag list instead of per call
- [x] Build the parser's per-line regexes once instead of per line
- [x] Build the `Util.balance` scanner regex once per pair instead of per call
- [x] Match interpolation without a MatchData and hoist the parser's lookup tables

## How to review

**One commit per optimization, in sequence — please review commit by commit.** Each commit message carries the motivation, the equivalence argument and the measured result for that change alone, and each commit is green in isolation, so any item can be reverted independently.

## Method

- 3 alternated rounds (`main` ↔ branch) per benchmark, median, `benchmark-ips`;
  both sides pinned to the same `Gemfile.lock` (temple 0.10.4), Ruby 4.0.2.
- Render allocations counted with `GC.stat(:total_allocated_objects)`.
- Rendered HTML of all 21 `benchmark/` templates diffed against `main`: byte-identical.
- Full suite green, including haml-spec: 1212 runs, 0 failures (49 tests added).

## Results

Control templates (untouched code paths) stay within ±1.6%; erubi in the slim benchmark, same gem on both sides, sits at +0.0%.

| benchmark                                      |         main |       branch |           Δ |
| ---------------------------------------------- | -----------: | -----------: | ----------: |
| render `dynamic_attributes/common_attribute`   |     556k ips |     891k ips |    **+60%** |
| render `data:`/`aria:` templates (3)           | 173–195k ips | 192–221k ips | **+11–13%** |
| render `dynamic_merger/hello` (interpolation)  |     276k ips |     307k ips |    **+11%** |
| render `slim/view`                             |     679k ips |     706k ips |     **+4%** |
| compile `etc/real_sample` (filters)            |     43.2 ips |     48.3 ips |    **+12%** |
| slim benchmark, haml row (erubi control +0.0%) |     661k ips |     690k ips |   **+4.3%** |

Allocations per render only go down (e.g. `data_attribute` 40 → 33, `etc/attribute_builder` 55 → 48); no template allocates more. The `escape_html_safe` gain doesn't show above because the bench doesn't load ActiveSupport — measured separately with AS loaded: +5.6% to +33%.

## Known trade-off and behavior notes

- `etc/string_interpolation` renders −5%: without the redundant `to_s`, `escape_html` now receives non-String values raw and pays C-side conversion. A `String === v` guard was evaluated and rejected (taxes the common path); candidate for a follow-up.
- Mutually-cyclic `data:` hashes now raise `SystemStackError` like every other cycle already did on `main`; the old deep `Hash#==` caught one narrow shape by accident. Directly self-referencing hashes keep working.
- `BOOLEAN_ATTRIBUTES` stays a public mutable Array; the derived Set rebuilds on size change, so same-size in-place edits and non-String elements (which the old splat matched via `===`) are no longer picked up — neither is documented usage.
- Fixed on the way: rendering `~` over `<pre>` content with only `haml/engine` required raised `NameError` (`Haml::Helpers` was never loaded).

### DIFF
--- a/lib/haml/attribute_builder.rb
+++ b/lib/haml/attribute_builder.rb
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
+require 'set'
 require 'haml/object_ref'
 
 module Haml::AttributeBuilder
+  @boolean_attributes = nil
+  @boolean_attributes_size = nil
+
   class << self
     def build(escape_attrs, quote, format, object_ref, *hashes)
       hashes << Haml::ObjectRef.parse(object_ref) if object_ref
@@ -19,10 +23,12 @@ module Haml::AttributeBuilder
           buf << build_data(escape_attrs, quote, format, *hash[key])
         when 'aria'
           buf << build_aria(escape_attrs, quote, format, *hash[key])
-        when *Haml::BOOLEAN_ATTRIBUTES, /\Adata-/, /\Aaria-/
-          build_boolean!(escape_attrs, quote, format, buf, key, hash[key])
         else
-          buf << " #{key}=#{quote}#{escape_html(escape_attrs, hash[key].to_s)}#{quote}"
+          if boolean_attribute?(key)
+            build_boolean!(escape_attrs, quote, format, buf, key, hash[key])
+          else
+            buf << " #{key}=#{quote}#{escape_html(escape_attrs, hash[key].to_s)}#{quote}"
+          end
         end
       end
       buf.join
@@ -71,6 +77,22 @@ module Haml::AttributeBuilder
     end
 
     private
+
+    # Haml::BOOLEAN_ATTRIBUTES is public API and documented as mutable, so the Set has to
+    # be derived lazily and rebuilt when the Array changes. The guard is the Array's size:
+    # an in-place mutation that keeps the size (BOOLEAN_ATTRIBUTES[0] = 'x') is not picked
+    # up, but adding attributes -- the documented usage -- always changes it. The Set is
+    # published before the size so a concurrent reader either sees the old size and
+    # rebuilds, or sees the new size with the new Set already visible.
+    def boolean_attribute?(key)
+      attributes = Haml::BOOLEAN_ATTRIBUTES
+      size = attributes.size
+      if @boolean_attributes_size != size
+        @boolean_attributes = Set.new(attributes)
+        @boolean_attributes_size = size
+      end
+      @boolean_attributes.include?(key) || key.start_with?('data-', 'aria-')
+    end
 
     def build_data_attribute(key, escape_attrs, quote, format, *hashes)
       attrs = []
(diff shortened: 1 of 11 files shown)


wilfison:
@k0kubun 

Happy to make any adjustments you think are needed — I can split, rework or drop any of the commits independently, since each one stands on its own. This is also the first, low-risk slice of a larger profiling effort: if this lands well, I have complementary follow-ups staged on top of it (compile-time resolution of static multi-line attribute hashes and a leaner `AttributeBuilder`, among others) that I'd be glad to bring as separate PRs.

k0kubun:
I want you to break up this PR, not to ten, but to four.

1. AttributeBuilder changes: https://github.com/haml/haml/pull/1215/changes/dd5dcef5b64c606e1bd0b3fa339b3baa2e2937af, https://github.com/haml/haml/pull/1215/changes/a9508ace27a2591518f88889c8e4cc4e79ae5f23, https://github.com/haml/haml/pull/1215/changes/f359601c27b1f5b5f47b6fd3aec4fd084981aea7
2. `html_safe` change: https://github.com/haml/haml/pull/1215/changes/080d5c3847983c50fd6d1e17910ec50d97d6d3d9
3. `find_and_preserve` change: https://github.com/haml/haml/pull/1215/changes/fa9e6699bba0237462acbcd5a8266b89efe2f84b
4. Other compiler/parser changes: https://github.com/haml/haml/pull/1215/changes/f1cd2fc75296f558dfcc1fd53845419585332160, https://github.com/haml/haml/pull/1215/changes/4e12e8a418770b19146ea9cd0f04801aa7789316, https://github.com/haml/haml/pull/1215/changes/3b83b3edf41aa5dd8c742bd51d6ee42d4a4a2d35, https://github.com/haml/haml/pull/1215/changes/a180714574d2b3c1848e44537ead55dc8956c5ec, https://github.com/haml/haml/pull/1215/changes/5e0946be52ad3d00110428ed1c10b3d45531fe64

I think I can merge (1) and (4) soon, but I want to think about (2) and (3) separately.

Also, I'd appreciate it if you can stop replying me in PRs that I haven't reviewed yet. It's not like I have bandwidth to review 800-line PRs every day. I occasionally come to this project and maintain it when I have time to do so, and I want to keep it that way.

k0kubun:
If it wasn't clear, I asked you to create four PRs at once, not in serial. My understanding is that these four don't depend on each other, so it should be possible. I think it will require less time (fewer round-trips, at the very least) for both of us.

PR #1216, opened 2026-08-11
Cut per-render overhead in AttributeBuilder

wilfison:
## Changes

One commit each:

- Classify boolean attributes with a Set instead of a `case`/`when` splat
- Look up Symbol attribute keys with `Symbol#name` instead of `to_s`
- Flatten nested attributes without `Hash#==` or an unconditional `tr`
- Derive the boolean attribute Set once instead of invalidating it per call — from your review

Each commit message carries the motivation, the equivalence argument and the measurement for that change alone, and each commit is green in isolation, so any one of them can be dropped independently.

## Results

| render                                 |     main |   branch |          Δ |
| -------------------------------------- | -------: | -------: | ---------: |
| `dynamic_attributes/common_attribute`  | 556k ips | 911k ips | **+63.8%** |
| `dynamic_attributes/data_attribute`    | 173k ips | 200k ips | **+15.9%** |
| `data_attribute`                       | 199k ips | 223k ips | **+12.3%** |
| `dynamic_attributes/boolean_attribute` | 552k ips | 619k ips | **+12.2%** |
| `etc/attribute_builder`                | 167k ips | 184k ips | **+10.5%** |

Those five are the templates that reach the changed code. The other 16 land between −2.3% and +1.2%, which is the noise floor of this bench: compiled Ruby is byte-identical throughout, so nothing outside the attribute path has anything to gain.

Allocations per render are the harder evidence, being counted rather than timed:

| template                               | main | branch |
| -------------------------------------- | ---: | -----: |
| `dynamic_attributes/data_attribute`    |   48 |     40 |
| `data_attribute`                       |   40 |     33 |
| `etc/attribute_builder`                |   55 |     48 |
| `dynamic_attributes/boolean_attribute` |   21 |     17 |
| `dynamic_attributes/common_attribute`  |   15 |     12 |

The other 14 templates allocate exactly the same; none allocates more.

Parity: compiled Ruby and rendered HTML byte-identical across all 21 `benchmark/` templates. Suite goes from 1156 to 1171 runs, 0 failures, with 2 previously skipped tests reactivated (199 → 197 skips).

## Behaviour notes

Three accepted changes, all in undocumented territory. Happy to guard any of them if you'd rather keep it working.

- The Set derived from `Haml::BOOLEAN_ATTRIBUTES` is built on first use and never rebuilt. Pushing to the list at boot, before anything renders, works as before. A list changed later is still honoured by the compile time path, which switches on the attribute key, so `%div{ custom: value }` keeps working; the one shape that stops seeing a late change is a fully dynamic hash, `%div{ hash }`, whose keys are only known at render. `old_attribute_test.rb` covers that shape, and both `with_custom_attributes` helpers now reset the memo so it stays covered.
- A non-String element pushed onto `BOOLEAN_ATTRIBUTES` no longer matches. The old splat tested each element with `===`, so a Regexp there worked as a pattern by accident. REFERENCE.md describes the constant as a list of attribute names.
- Mutually-cyclic `data:` hashes now raise `SystemStackError`. The general mutual cycle (`a[:x] = b`, `b[:y] = a`) already did on `main`; the deep `Hash#==` only caught the narrow `a[:data] = b`, `b[:data] = a` shape by accident, and that one rendered `<div></div>`. A directly self-referencing hash, the shape the suite and the docs describe, is unchanged.

The two self-reference tests in `test/haml/engine/attributes_test.rb` were skipped and used `assert_haml`, which compares a template against itself and asserts nothing. They are now `assert_render` and reactivated.

### DIFF
--- a/lib/haml/attribute_builder.rb
+++ b/lib/haml/attribute_builder.rb
@@ -141,7 +141,8 @@ module Haml::AttributeBuilder
           raise ArgumentError, "Non-hash object is given to attributes!"
         end
         hash.each do |key, value|
-          key = key.to_s
+          # Symbol#name hands back the interned String; to_s allocates one per render.
+          key = key.is_a?(Symbol) ? key.name : key.to_s
           case key
           when 'id', 'class', 'data', 'aria'
             merged[key] ||= []
--- a/lib/haml/attribute_builder.rb
+++ b/lib/haml/attribute_builder.rb
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
+require 'set'
 require 'haml/object_ref'
 
 module Haml::AttributeBuilder
+  @boolean_attributes = nil
+  @boolean_attributes_size = nil
+
   class << self
     def build(escape_attrs, quote, format, object_ref, *hashes)
       hashes << Haml::ObjectRef.parse(object_ref) if object_ref
@@ -19,10 +23,12 @@ module Haml::AttributeBuilder
           buf << build_data(escape_attrs, quote, format, *hash[key])
         when 'aria'
           buf << build_aria(escape_attrs, quote, format, *hash[key])
-        when *Haml::BOOLEAN_ATTRIBUTES, /\Adata-/, /\Aaria-/
-          build_boolean!(escape_attrs, quote, format, buf, key, hash[key])
         else
-          buf << " #{key}=#{quote}#{escape_html(escape_attrs, hash[key].to_s)}#{quote}"
+          if boolean_attribute?(key)
+            build_boolean!(escape_attrs, quote, format, buf, key, hash[key])
+          else
+            buf << " #{key}=#{quote}#{escape_html(escape_attrs, hash[key].to_s)}#{quote}"
+          end
         end
       end
       buf.join
@@ -71,6 +77,19 @@ module Haml::AttributeBuilder
     end
 
     private
+
+    # BOOLEAN_ATTRIBUTES is mutable public API, so the Set is rebuilt whenever the Array's
+    # size changes. The Set is published before the size, so a concurrent reader either
+    # rebuilds or sees the new Set.
+    def boolean_attribute?(key)
+      attributes = Haml::BOOLEAN_ATTRIBUTES
+      size = attributes.size
+      if @boolean_attributes_size != size
+        @boolean_attributes = Set.new(attributes)
+        @boolean_attributes_size = size
+      end
+      @boolean_attributes.include?(key) || key.start_with?('data-', 'aria-')
+    end
 
     def build_data_attribute(key, escape_attrs, quote, format, *hashes)
       attrs = []
--- a/lib/haml/attribute_builder.rb
+++ b/lib/haml/attribute_builder.rb
@@ -1,10 +1,7 @@
 # frozen_string_literal: true
-require 'set'
 require 'haml/object_ref'
 
 module Haml::AttributeBuilder
-  @boolean_attributes = nil
-
   class << self
     def build(escape_attrs, quote, format, object_ref, *hashes)
       hashes << Haml::ObjectRef.parse(object_ref) if object_ref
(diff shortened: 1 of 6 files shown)


lib/haml/attribute_builder.rb
```diff
@@ -72,6 +78,19 @@ def build_aria(escape_attrs, quote, format, *hashes)
 
     private
 
+    # BOOLEAN_ATTRIBUTES is mutable public API, so the Set is rebuilt whenever the Array's
+    # size changes. The Set is published before the size, so a concurrent reader either
+    # rebuilds or sees the new Set.
```
k0kubun: This seems rather suboptimal for most users, introducing an extra check that runs every time. `Haml::BOOLEAN_ATTRIBUTES` is meant to be modified only for very rare use cases and not supposed to be used by most people; otherwise the official default should be changed instead.

Can you drop the `@boolean_atrributes` and `@boolean_attributes_size`-based optimization?

wilfison: Done in `9807486`.

The Set is now derived on first use and never rebuilt. It stays lazy rather than becoming a load time constant because `attribute_compiler.rb` defines `BOOLEAN_ATTRIBUTES` and requires `attribute_builder.rb`, so the constant does not exist yet when this file loads.

One consequence worth your call: a list changed after the first build is now only seen by the compile time path, which switches on the attribute key. `%div{ custom: value }` still works, since the key is known at compile time; the shape that loses it is a fully dynamic `%div{ hash }`. `old_attribute_test.rb` has exactly that line, so both `with_custom_attributes` helpers now reset the memo to keep it covered. If you would rather not have tests reaching for an ivar, I can drop that line and record the behaviour change instead.

Worth about 3% on top of what the Set already gained, on `dynamic_attributes/common_attribute`.

k0kubun: Alright, now I see what you wanted out of it.

Can you just turn `Haml::BOOLEAN_ATTRIBUTES` itself into a `Set`? As soon as this method gets executed once, any modification to `Haml::BOOLEAN_ATTRIBUTES` would be ignored only in this place, so this implementation is unsound.

This was originally supposed to be used with https://github.com/hamlit/hamlit-boolean_attributes/blob/master/lib/hamlit/boolean_attributes.rb, which uses `#<<`, so that should work with `Set` as well.

wilfison: Done!

I also swapped the compile time `when *BOOLEAN_ATTRIBUTES` for the same `include?` check the runtime uses: splatting a Set rebuilds an Array on every key, and this way both paths classify identically. Happy to revert that hunk if you would rather keep the diff to the constant itself.

Two API notes, both worth a changelog line if you agree. `Set` has no `push`, so that call becomes `<<` or `merge`, hamlit-boolean_attributes already uses `<<`. And a Regexp in the list no longer works as a pattern, since the old splat compared with `===`.

lib/haml/attribute_builder.rb
```diff
@@ -98,14 +104,18 @@ def flatten_attributes(attributes)
...
         case value
-        when attributes
         when Hash
           flatten_attributes(value).each do |k, v|
             if k.nil?
               flattened[key] = v
             else
-              flattened["#{key}-#{k.to_s.tr('_', '-')}"] = v
+              k = k.is_a?(Symbol) ? k.name : k.to_s
+              k = k.tr('_', '-') if k.include?('_')
```
k0kubun, line 117: Are you sure this `if k.include?('_')` helps? This seems like an unnecessary double scan.

```suggestion
              k = k.tr('_', '-')
```

wilfison: It does, because `tr` allocates a new String unconditionally even when nothing is replaced, while `include?` only scans. The trade isn't scan-vs-scan, it's scan vs scan-plus-allocate, and the result is discarded right away by `"#{key}-#{k}"`.

Measured on Ruby 4.0.2 with the guard removed:

- `benchmark/dynamic_attributes/data_attribute.haml`: 40 -> 44 allocations per render,
  181.5k -> 173.4k i/s (-4.4%)
- `benchmark/etc/attribute_builder.haml`: 48 -> 52 allocations per render,
  170.4k -> 165.7k i/s (-2.9%)

In that first template the line runs for `id`, `name`, `user-id`, `user-name`, `book_id`; only `book_id` converts, so 4 allocations are spared, matching the delta exactly. Note that `user-id`/`user-name` arrive already hyphenated from the inner pass, so deeper nesting favors the guard more.

On the line alone: 2.03x faster for keys without an underscore, 1.30x slower for keys with one, 1.18x faster on a 50/50 mix. It would only lose if most nested keys carried underscores.

k0kubun: makes sense. `tr!` isn't particularly handy in this situation either. It seems fine that we optimize for the most reasonable case that you don't use `_`.

k0kubun (approved):

Issue #1084, opened 2022-09-22
haml6 & string interpolation escapes


jdelStrother:
Thanks for haml6! But I'm puzzled by a change we're seeing with it in our Rails 7.0.3 app.  A template like this:

```haml
%div
  %p= link_to "link1", "/"
  %p click #{link_to "link2", "/"}
```

Generates this html in haml 5.2.2:

```html
<div>
<p><a href="/">link1</a></p>
<p>click <a href="/">link2</a></p>
</div>
```

But generates this html in haml 6.0.0:

```html
<div>
<p><a href="/">link1</a></p>
<p>click &lt;a href=&quot;/&quot;&gt;link2&lt;/a&gt;</p>
</div>
```

Should we still be able to use `foo #{html_safe_string} bar` without `html_safe_string` getting over-escaped?

jdelStrother:
Possibly in haml/util.rb it should be something like this?

```diff
    def unescape_interpolation(str, escape_html = nil)
      res = ''.dup
      rest = Haml::Util.handle_interpolation str.dump do |scan|
        escapes = (scan[2].size - 1) / 2
        char = scan[3] # '{', '@' or '$'
        res << scan.matched[0...-3 - escapes]
        if escapes % 2 == 1
          res << "\##{char}"
        else
          interpolated = if char == '{'
            balance(scan, ?{, ?}, 1)[0][0...-1]
          else
            scan.scan(/\w+/)
          end
          content = eval("\"#{interpolated}\"")
          content = "#{char}#{content}" if char == '@' || char == '$'
-          content = "CGI.escapeHTML((#{content}).to_s)" if escape_html
+          content = "Haml::Util.escape_html_safe((#{content}).to_s)" if escape_html

          res << "\#{#{content}}"
        end
      end
      res + rest
    end
```

k0kubun:
I'll need more time to work on a better solution, but I accepted your patch for now. Thank you!

k0kubun:
Released as v6.0.1.

PR #1217, opened 2026-08-14
Pick the html_safe escaper once at load instead of per value

wilfison:
## Change

`Haml::Util.escape_html_safe` ran `respond_to?(:html_safe?)` on every escaped value, but the
answer never depended on the value. The first line is `html = html.to_s`, so the receiver is
always a String by then, and ActiveSupport defines `html_safe?` on `Object` — it is a
process-global property, checked per call. The escaper is now chosen once, at load, on the same
predicate Temple uses to default `use_html_safe`.

This is not only the Rails path: `Haml::Escape` builds the escaper name from
`options[:use_html_safe]`, so every `=` goes through `escape_html_safe` whenever ActiveSupport
is loaded, Tilt included.

The guarded variant stays for the no-ActiveSupport case, because `unescape_interpolation` bakes
the call into a String literal (`util.rb:221`), outside the temple tree, so it ignores
`use_html_safe` entirely. Both branches are correct on their own, and the unguarded one is only
chosen when ActiveSupport is already loaded, which cannot be undone.

## Benchmark

The call itself, both variants defined side by side in one process (`benchmark-ips`, 2s warmup,
5s measured):

| value        |     before |      after |          |
| ------------ | ---------: | ---------: | -------: |
| plain String |  7.50M ips |  8.95M ips | **1.19x** |
| SafeBuffer   | 15.07M ips | 20.30M ips | **1.35x** |
| Integer      | 10.35M ips | 12.94M ips | **1.25x** |

Render, median of 5 alternating rounds over every `benchmark/` template, ActiveSupport loaded,
with the count of `escape_html_safe` calls in each template's compiled Ruby:

| template                   | calls |      main |    branch |          Δ |
| -------------------------- | ----: | --------: | --------: | ---------: |
| `dynamic_merger/hello`     |    50 |  246k ips |  311k ips | **+26.1%** |
| `plain`                    |     2 | 2.69M ips | 3.03M ips | **+12.7%** |
| `script`                   |     1 | 3.65M ips | 4.03M ips | **+10.6%** |
| `etc/string_interpolation` |     1 | 2.65M ips | 2.88M ips |  **+8.8%** |
| `slim/view`                |     3 |  600k ips |  648k ips |  **+8.0%** |
| `etc/real_sample`          |    59 |   37k ips |   40k ips |  **+7.6%** |

Those six are the templates that reach the code. The other 15 compile to zero calls and are the
control: they land between +0.0% and +10.3%, and that upper end is noise — 13 of the 15 sit at
or below +4.2%, and the two above it are also the two noisiest, at 9.8% and 15.5% per-side
dispersion against 2.4% for `hello`. `etc/real_sample` gains least per call because escaping is
a small share of what it renders.

Rendered HTML byte-identical across all 21 templates. Suite green: 1161 runs, 0 failures.

### DIFF
--- a/lib/haml/util.rb
+++ b/lib/haml/util.rb
@@ -26,10 +26,9 @@ module Haml
       end
     end
 
-    # TODO: Remove unescape_interpolation's workaround and get rid of `respond_to?`.
     def self.escape_html_safe(html)
       html = html.to_s
-      (html.respond_to?(:html_safe?) && html.html_safe?) ? html : escape_html(html)
+      html.html_safe? ? html : escape_html(html)
     end
 
     # Silence all output to STDERR within a block.
@@ -202,7 +201,7 @@ MSG
       /#[\{$@]/ === str
     end
 
-    def unescape_interpolation(str, escape_html = nil)
+    def unescape_interpolation(str)
       res = ''.dup
       rest = Haml::Util.handle_interpolation str.dump do |scan|
         escapes = (scan[2].size - 1) / 2
@@ -218,7 +217,6 @@ MSG
           end
           content = eval("\"#{interpolated}\"")
           content = "#{char}#{content}" if char == '@' || char == '$'
-          content = "Haml::Util.escape_html_safe((#{content}).to_s)" if escape_html
 
           res << "\#{#{content}}"
         end
--- a/lib/haml/parser.rb
+++ b/lib/haml/parser.rb
@@ -444,7 +444,8 @@ module Haml
         end
       else
         if Util.contains_interpolation?(value)
-          value = Util.unescape_interpolation(value, escape_html)
+          value = Util.unescape_interpolation(value)
+          escape_interpolation = true if escape_html
           parse = true
           escape_html = false
         end
--- a/test/haml/optimization_test.rb
+++ b/test/haml/optimization_test.rb
@@ -215,4 +215,62 @@ describe 'optimization' do
       assert_operator saved, :>, RUNS / 2
     end
   end
+
+  # test_helper loads rails before haml, so use_html_safe defaults to true here.
+  describe 'html safe escaping' do
+    class ::TosSafeBufferObject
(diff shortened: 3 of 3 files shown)


lib/haml/util.rb
```diff
@@ -26,10 +26,18 @@ def self.escape_html(html)
       end
     end
 
-    # TODO: Remove unescape_interpolation's workaround and get rid of `respond_to?`.
```
k0kubun, line 29: Can you solve this TODO instead? Branching the definition is not the right resolution to this problem. There's no point in using `escape_html_safe` if you don't use Rails. So, this PR should just get rid of `html.respond_to?(:html_safe?) && ` and make the rest work without leaving the fallback.

wilfison: Done! The TODO is solved without branching the definition.

`Parser#tag` bakes the escaper into the string literal for interpolated tag text, which 08098f1b pointed at `escape_html_safe` for #1084:

```ruby
%p hello #{name}   # => "hello #{Haml::Util.escape_html_safe((name).to_s)}"
```

That literal never reaches `Haml::Escape`, so it can't see `use_html_safe`, hence the `respond_to?` guard added in the same commit. The other interpolation branches (`=`, `&`/`!`) instead set `escape_interpolation` and let `TagCompiler#compile_interpolated_plain` emit an `:escape` node per value. Matching them puts `Haml::Escape` in charge, and it names `escape_html_safe` only when `use_html_safe` is on, so the guard is unreachable and gone, along with `unescape_interpolation`'s now unused second parameter.

`escape_html_safe` still asks `html_safe?` of `to_s`, so kaminari-style objects are unaffected and #1084 stays fixed. `%p hello #{name}` now honors `use_html_safe: false`. Tests cover both directions of the option, plus a subprocess without ActiveSupport asserting `escape_html_safe` is never compiled in.

k0kubun (approved):

PR #1219, opened 2026-08-14
Keep Prism-derived fragments in the source encoding

### LINKED ISSUE #1218: Encoding::CompatibilityError when compiling ASCII-8BIT template source with non-ASCII characters (regression from 7.2.2 in 7.3.0)
## Summary

haml 7.3.0 raises `Encoding::CompatibilityError` when compiling a template whose source string is in ASCII-8BIT and contains non-ASCII characters both in plain text and in embedded Ruby code (e.g. a string literal in an attribute hash). 7.2.2 compiles the same input fine, so this is a regression, most likely from the Ripper -> Prism migration (#1214).

## Minimal reproduction (no Rails required)

```ruby
# Usage:
#   HAML_VERSION=7.2.2 ruby repro.rb  # => <p title="🍣">🍺</p>
#   HAML_VERSION=7.3.0 ruby repro.rb  # => Encoding::CompatibilityError
require 'bundler/inline'

gemfile(true, quiet: true) do
  source 'https://rubygems.org'
  gem 'haml', ENV.fetch('HAML_VERSION', '7.3.0')
end

require 'haml'

# Non-ASCII characters in BOTH plain text and a Ruby string literal,
# with the source in ASCII-8BIT (as returned by File.binread).
source = "%p{title: '🍣'} 🍺\n".dup.force_encoding(Encoding::BINARY)

puts "haml #{Haml::VERSION}"
puts Haml::Template.new { source }.render
```

Output on 7.2.2:

```
haml 7.2.2
<p title="🍣">🍺</p>
```

Output on 7.3.0:

```
temple-0.10.6/lib/temple/filters/static_merger.rb:23:in 'block in Temple::Filters::StaticMerger#on_multi': incompatible character encodings: UTF-8 and BINARY (ASCII-8BIT) (Encoding::CompatibilityError)
  from temple-0.10.6/lib/temple/filters/static_merger.rb:20:in 'Array#each'
  from temple-0.10.6/lib/temple/filters/static_merger.rb:20:in 'Temple::Filters::StaticMerger#on_multi'
  from temple-0.10.6/lib/temple/mixins/dispatcher.rb:69:in 'Temple::Filters::StaticMerger#dispatcher'
  ...
  from haml-7.3.0/lib/haml/template.rb:16:in 'Haml::TemplateExtension#compile'
```

Both non-ASCII pieces are required to trigger it:

- `%p 🍺` alone: OK
- `%p= '🍣'` alone: OK
- non-ASCII in plain text **and** in a Ruby string literal: crash

Depending on the template structure the crash site can also be `Temple::Generator#on_multi` instead of `StaticMerger` — in both cases it is the concatenation of adjacent static fragments with mixed encodings.

## What seems to be happening

Fragments that go through Prism come back re-encoded (UTF-8), while fragments taken directly from the source string keep the source's encoding (ASCII-8BIT). When Temple later joins adjacent static fragments, the mix blows up if both sides contain non-ASCII bytes. 7.2.2 (Ripper-based) kept the generated code uniformly in the source encoding, so this never surfaced.

## Why this hits real Rails apps in production

Rails' `ActionView::DependencyTracker::RubyTracker` compiles templates to extract render dependencies, and it passes `template.source` to the handler **without re-encoding** ([ruby_tracker.rb:31](https://github.com/rails/rails/blob/v8.1.3.1/actionview/lib/action_view/dependency_tracker/ruby_tracker.rb#L31)). Since `FileTemplate` sources are read with `File.binread`, the handler receives ASCII-8BIT bytes.

This path runs whenever `config.action_controller.perform_caching = true` and a template uses the `cache` helper (via `ActionView::Digestor`). So for any app with non-ASCII templates (e.g. Japanese) plus fragment caching:

- production 500s with `ActionView::Template::Error: incompatible character encodings`
- while CI stays green, because the Rails default for the test environment is `perform_caching = false`, so the Digestor path is never exercised by tests

## Workaround we are using

Normalizing the source encoding at the entrance of `Haml::RailsTemplate#call`:

```ruby
# config/initializers/haml_binary_template_source_patch.rb
if Gem.loaded_specs['haml'].version >= Gem::Version.new('7.3.0')
  module Haml
    module BinaryTemplateSource
      def call(template, source = nil)
        source ||= template.source
        if source.is_a?(String) && source.encoding == Encoding::BINARY
          utf8 = source.dup.force_encoding(Encoding::UTF_8)
          source = utf8 if utf8.valid_encoding?
        end

        super(template, source)
      end
    end
  end

  Haml::RailsTemplate.prepend(Haml::BinaryTemplateSource)
end
```

Something equivalent inside haml itself (normalizing the source encoding before parsing, or forcing the Prism-derived fragments back to the source encoding as 7.2.2 effectively did) would let us drop the patch.

## Environment

- haml 7.3.0 (regression from 7.2.2)
- temple 0.10.6
- Ruby 4.0.6
- Rails 8.1.3.1 (for the production trigger path; the minimal repro doesn't need Rails)

Thanks for maintaining haml!

fukayatsu:
Fixes #1218.

## Problem

Since the Ripper -> Prism migration (#1214), compiling a template whose source string is ASCII-8BIT raises `Encoding::CompatibilityError` when non-ASCII characters appear both in plain text and in embedded Ruby code:

```ruby
source = "%p{title: '🍣'} 🍺\n".dup.force_encoding(Encoding::BINARY)
Haml::Template.new { source }.render
# 7.2.2 => <p title="🍣">🍺</p>
# 7.3.0 => Encoding::CompatibilityError: incompatible character encodings: UTF-8 and BINARY (ASCII-8BIT)
```

This hits Rails apps in production: `ActionView::DependencyTracker` compiles `template.source` as `File.binread` returned it (ASCII-8BIT), so any app with non-ASCII templates plus fragment caching gets 500s once `perform_caching` is on, while CI stays green because the test environment defaults to `perform_caching = false`. Details in #1218.

## Cause

Prism tags what it returns with the encoding it parsed the code under — UTF-8 for a BINARY source. That applies both to node/location slices (`Prism::Source.for` re-tags a BINARY source holding valid multibyte UTF-8 as UTF-8) and to `StringNode#unescaped`. Meanwhile the fragments Haml slices out of the source itself (plain text, `byteslice`d interpolation bodies) keep the source encoding. The compiled template mixes the two kinds of fragments, and Temple's `StaticMerger` / `Generator#on_multi` join adjacent static fragments, which raises once both sides hold non-ASCII bytes. Ripper reported everything in the source encoding, which is why 7.2.2 was fine.

## Fix

Force everything taken from Prism back to the source encoding at the two places where Prism-derived text enters the compiled output, `StringSplitter` and `AttributeParser`, restoring the Ripper-era invariant that all fragments share the source encoding. This is a byte-preserving re-tag (`dup.force_encoding`), a no-op whenever the source is already UTF-8.

Tests cover the three Prism-derived paths (`unescaped`, `EmbeddedVariableNode`, attribute keys/values) plus an end-to-end render of the reproduction above; the engine test fails with `Encoding::CompatibilityError` before this change. The full suite passes locally (1161 runs, 0 failures, 0 errors).

Thanks for maintaining haml!

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Haml Changelog
 
+## Unreleased
+
+* Keep Prism-derived fragments in the source encoding, fixing `Encoding::CompatibilityError` on an ASCII-8BIT template source with non-ASCII characters (regression in 7.3.0) https://github.com/haml/haml/issues/1218
+
 ## 7.3.0
 
 * Replace Ripper with Prism https://github.com/haml/haml/pull/1214
--- a/lib/haml/attribute_parser.rb
+++ b/lib/haml/attribute_parser.rb
@@ -32,7 +32,7 @@ module Haml
         key = static_key(element.key)
         return if key.nil?
 
-        hash[key] = value_source(element.value)
+        hash[in_source_encoding(key, exp)] = in_source_encoding(value_source(element.value), exp)
       end
       hash
     end
@@ -73,6 +73,16 @@ module Haml
       return '' if value.is_a?(Prism::ImplicitNode)
 
       value.slice
+    end
+
+    # Prism tags its slices with the encoding it parsed the source under (UTF-8 for a
+    # BINARY source), while the rest of the compiled template stays in the source
+    # encoding. Mixing the two raises Encoding::CompatibilityError once both sides hold
+    # non-ASCII bytes, so bring everything back to the source encoding.
+    def in_source_encoding(string, source)
+      return string if string.encoding == source.encoding
+
+      string.dup.force_encoding(source.encoding)
     end
   end
 end
--- a/lib/haml/string_splitter.rb
+++ b/lib/haml/string_splitter.rb
@@ -10,7 +10,7 @@ module Haml
       def compile(code, node: RubyExpression.string_literal_node(code))
         case node
         when Prism::StringNode
-          node.unescaped.empty? ? [] : [[:static, node.unescaped]]
+          node.unescaped.empty? ? [] : [[:static, in_source_encoding(node.unescaped, code)]]
         when Prism::InterpolatedStringNode
           compile_parts(node.parts, code)
         else
@@ -33,12 +33,12 @@ module Haml
             case part
             when Prism::StringNode
               content = part.unescaped
-              exps << [:static, content] unless content.empty?
+              exps << [:static, in_source_encoding(content, code)] unless content.empty?
             when Prism::EmbeddedStatementsNode
               embedded = embedded_source(part, code)
               exps << [:dynamic, embedded] unless embedded.empty?
             when Prism::EmbeddedVariableNode
-              exps << [:dynamic, part.variable.slice]
+              exps << [:dynamic, in_source_encoding(part.variable.slice, code)]
             else
               # Prism allows more part types than a string literal can currently hold. Dropping
               # one would silently lose content, so fail instead of rendering something wrong.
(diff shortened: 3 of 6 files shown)


k0kubun (approved):

fukayatsu:
Thanks!

PR #1220, opened 2026-08-14
Build the preserve regex once per tag list instead of per call

wilfison:
## Summary

`find_and_preserve` rebuilt its regex on every call: `Regexp.escape` per tag, a join, and a
fresh `Regexp` interpolated from the result. Two copies did it, one in `ScriptCompiler` and
one in `RailsHelpers`, and they had already drifted apart. They are now one implementation in
`Haml::Helpers`, behind a copy-on-write cache keyed by tag list.

Split into small commits, each green on its own, each carrying its own rationale in the
commit message.

## User visible changes

- **A `NameError` is fixed.** `haml/engine.rb` never required `haml/helpers`, so
  `require 'haml/engine'` alone, rendering `~` over content holding a `<pre>`, crashed at
  `script_compiler.rb:14`. Subprocess test included.
- **An empty tag no longer joins the pattern**, where it used to match a literal `<>`. The
  two builders also disagreed about a leading one.
- **`Haml::RailsHelpers::DEFAULT_PRESERVE_TAGS` is no longer declared there.** It still
  resolves, to the same object, through `include Helpers`; only `constants(false)` can tell.
- **`preserve` matches a newline with a String, not a Regexp.** Equivalent except on a broken
  coderange, where the next line, `delete!("\r")`, raises the same error anyway.

## Benchmarks

Ruby 4.0.2, old implementation reproduced verbatim beside the new:

| | before | after | |
|---|---:|---:|---:|
| `find_and_preserve`, no match | 201k ips | 3519k ips | 17.5x |
| `find_and_preserve`, one `<pre>` | 139k ips | 511k ips | 3.66x |
| render of `!~ x`, compiled shape | 140k ips | 512k ips | 3.64x |
| `find_and_preserve`, 20 lines | 76k ips | 147k ips | 1.95x |

The first row is the common case: `~` is often used defensively over content that holds no
preserved tag, and that used to pay full construction to then match nothing. The gain shrinks
as input grows, because then matching dominates. Allocations per call go 20 to 2 with no
match; a cached lookup allocates nothing.

The cache is a frozen Hash replaced on a miss, never mutated, so a racing writer on jruby or
truffleruby at worst rebuilds a regex. Rationale for the key and the size cap is in the
message of the commit that adds it.

### DIFF
--- a/lib/haml/helpers.rb
+++ b/lib/haml/helpers.rb
@@ -1,6 +1,24 @@
 # frozen_string_literal: true
 module Haml
   module Helpers
+    DEFAULT_PRESERVE_TAGS = %w[textarea pre code].freeze
+
+    # The two callers differ only in how the preserved content is marked, which is what
+    # the block is for; everything around it, the SafeBuffer workaround included, is here.
+    #
+    # @api private
+    def self.find_and_preserve(input, tags)
+      # An empty entry would add an alternative matching `<>`.
+      pattern = tags.reject(&:empty?).map { |tag| Regexp.escape(tag) }.join('|')
+      re = /<(#{pattern})([^>]*)>(.*?)(<\/\1>)/im
+      input.to_s.gsub(re) do |s|
+        s =~ re # Can't rely on $1, etc. existing since Rails' SafeBuffer#gsub is incompatible
+        name, attributes, content = $1, $2, $3
+        content = block_given? ? yield(content) : preserve(content)
+        "<#{name}#{attributes}>#{content}</#{name}>"
+      end
+    end
+
     def self.preserve(input)
       s = input.to_s.chomp("\n")
       s.gsub!("\n", '&#x000A;')
--- a/lib/haml/helpers.rb
+++ b/lib/haml/helpers.rb
@@ -1,63 +1,6 @@
 # frozen_string_literal: true
 module Haml
   module Helpers
-    # Referenced by name from the emitted Ruby, so every render hands preserve_regex the
-    # same object.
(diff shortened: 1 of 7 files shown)


lib/haml/helpers.rb
```diff
@@ -1,9 +1,66 @@
...
+        @preserve_regexes = cache.merge(key => regex).freeze
+        regex
+      end
+    end
+
+    # The two callers differ only in how the preserved content is marked, which is what
+    # the block is for; everything around it, the SafeBuffer workaround included, is here.
+    #
+    # @api private
+    def self.find_and_preserve(input, tags)
```
k0kubun: Can we avoid defining new methods on `Haml::Helpers` please? This is the most user-facing module in Haml and I don't want to add any new interface to it. Even the private constants scare me since it ends up being included in many places and it's very likely that they're actually visible when user-defined code is evaluated.

If you can't find an appropriate place to share some code, you may define a new module and define them there.

wilfison: Makes sense...

`Haml::Helpers` is back to what it was on main, except for the `gsub!("\n", ...)` inside the existing `preserve`.

The shared code now lives in `Haml::Preserver` (`lib/haml/preserver.rb`), a new module that is never included anywhere, so neither the methods nor the private constants can show up where user code is evaluated. `RailsHelpers` and the compiled-template runtime call into it, and `RailsHelpers::DEFAULT_PRESERVE_TAGS` stays where it was.

Also rebased onto main.

k0kubun (approved):

PR #1222, opened 2026-08-18
Compile a multi-line attribute hash statically

wilfison:
## Change

`%div{ class: "card", id: "main" }` compiles to one static String, but the same hash broken across two lines called `AttributeBuilder.build` on every render, because `AttributeParser` returned nil on any newline. The guard existed for line numbers: `ChildrenCompiler` advances its line count by the newlines in the attribute source, and the runtime path embeds that source verbatim. The static path now keeps the same contract by asking `TempleLineCounter` how many lines its temple spans and appending the `[:newline]` nodes still missing; a newline inside a value emitted verbatim is counted, not doubled. The count is only taken when the source has a newline, so a single-line hash compiles at the same cost as before. A hash holding a heredoc stays on the runtime, since a heredoc's slice does not include its body.

## Behaviour change

The values of a multi-line hash are evaluated on the tag's first line and in attribute order, exactly as a single-line hash already is. `__LINE__` inside such a value, and the backtrace of an exception raised there, point at the tag; the lines of the tag's content and of everything after it are unchanged (`line_number_test` updated from `b="2"` to `b="1"`). `Haml::AttributeParser.parse` now returns a Hash for a multi-line hash instead of nil.

## Results

Render on ruby 4.0.2, Tilt path, same template files on both sides:

| render | main | branch |
| --- | ---: | ---: |
| `benchmark/multiline_attribute.haml` (new, twin of `common_attribute`) | 305k ips | 2.70M ips (**8.9x**) |
| `benchmark/common_attribute.haml` (its single-line twin) | 2.81M ips | 2.76M ips |
| the 3-line hash of the report, alone | 420k ips | 8.75M ips (**20.8x**) |
| the same hash on one line, alone | 20.4M ips | 19.7M ips |

The remaining gap on the last pair is a Tilt-only artefact: a template that is a single static tag collapses to `_buf = "..."`, and the `[:newline]` nodes keep it from doing so; the Rails generator has no such shortcut, and any template with one more line does not take it either, hence the twins tying. The 21 existing benchmark templates compile to byte-identical Ruby. Suite at 1227 runs, 0 failures.

### DIFF
--- a/benchmark/multiline_attribute.haml
+++ b/benchmark/multiline_attribute.haml
@@ -1,0 +1,5 @@
+%a{ href: '&"\'<>',
+    title: 'x' }
+- href = '&"\'<>'
+%a{ href: href,
+    title: 'x' }
--- a/lib/haml/attribute_compiler.rb
+++ b/lib/haml/attribute_compiler.rb
@@ -3,6 +3,7 @@ require 'set'
 require 'haml/attribute_builder'
 require 'haml/attribute_parser'
 require 'haml/ruby_expression'
+require 'haml/temple_line_counter'
 
 module Haml
   # The set of boolean attributes. You may add custom attributes to this constant.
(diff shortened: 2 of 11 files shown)


k0kubun (approved):
> The values of a multi-line hash are evaluated on the tag's first line and in attribute order, exactly as a single-line hash already is. `__LINE__` inside such a value, and the backtrace of an exception raised there, point at the tag

This should be fixed at some point. Whether it's completely static or not, there should be a way to construct a Ruby code in a way that would still render `__LINE__` properly.

Though I would rather like to land this optimization first and leave it for a separate PR to fix it.

